Issue #2, opened 2021-09-01
GPG identity <> ETH address / ENS name


sebastinez:
## Abstract
This issue is for the debate on the linking of GPG identities with a ETH address and a registered ENS name.
It'll eventually allow for the web app and other services to verify through a cryptographic proof, that a commit was indeed signed by the corresponding `Person`.
It's heavily based on https://github.com/radicle-dev/radicle-link/pull/525 and https://github.com/radicle-dev/radicle-upstream/issues/965

## Requirements 
- The link must be verifiable with a cryptographic proof.
- Both directions should be verifiable: GPG -> Radicle and Radicle -> GPG
- The link and proof must be retrievable with just the GPG key fingerprint
- The GPG identity must be retrievable with just the Ethereum address.
- A command-line tool must be provided to create this link, and it should support the Ledger Nano S/X
- The proof should be published on IPFS or other decentralized storage
- The web client should integrate with this system and show the Radicle identities of project contributors who have signed their commits and linked their identities.

sebastinez:
Will close this issue for now, since we are moving to SSH and other attestation flows

PR #5, opened 2021-09-17
Proof creation

sebastinez:
This PR works on the creation of one-way proofs to link a GPG identity to a ETH address and vice versa.

It's still pretty much in alpha/beta so any feedback is welcome and specially from of a real cryptographer would not harm this PR.

In essence the new CLI allows a user using their GPG Long Key and a keystore file or a Ledger HW to sign two messages both showing that a specific GPG key corresponds to a specific ETH address, and for time relevance / revocation it injects a block hash to allow eventual future proofs to define which one came last.

This PR is based on https://github.com/radicle-dev/radicle-link/pull/557 and #2

### DIFF
--- a/proof-generator/USAGE
+++ b/proof-generator/USAGE
@@ -1,0 +1,20 @@
+Usage
+
+  proof-generator --gpg-key <string> --keystore <file> --rpc-url <url> --output <file> [<option>..] 
+
+Options
+
+    --gpg-key <string>           GPG key fingerprint
+    --keystore <file>            Path to keystore file
+    --ledger-hdpath <string>     HD Derivation path of Ledger HW
+    --rpc-url <url>              JSON-RPC URL of Ethereum node (eg. http://localhost:8545)
+    --output <file>              Path to where the proof should be stored
+    -v, --verbose                Verbose output
+    --help                       Show this message
+
+Example
+
+  proof-generator --gpg-key EB1729638209DCE61281F416504C9C1DE8C47EDF \
+             --rpc-url http://localhost:8545 \
+             --ledger-hdpath "m/44'/60'/0'/0/0" \
+             --output ./proof.json
--- a/proof-generator/src/lib.rs
+++ b/proof-generator/src/lib.rs
@@ -1,0 +1,177 @@
+use anyhow::anyhow;
+use coins_bip32::path::DerivationPath;
+use ethers::{
+    prelude::Signer,
+    providers::{Http, Middleware, Provider},
+    signers::{HDPath, Ledger},
+    types::{Signature, H256, H160},
+};
+use serde::{Deserialize, Serialize};
+use std::{
+    convert::TryFrom,
+    fmt::{Debug, Display},
+    fs,
+    io::Write,
+    path::PathBuf,
+    process::{Command, Stdio},
+    str,
+    borrow::Borrow,
+};
+
+/// The options allowed to be provided to the CLI
+#[derive(Debug, Clone)]
+pub struct Options {
+    /// GPG key to sign the proof.
+    pub gpg_key: String,
+    /// Output path of created proof
+    pub output: PathBuf,
+    /// RPC url
+    pub rpc_url: String,
+    /// Account derivation path when using a Ledger hardware wallet.
+    pub ledger_hdpath: Option<DerivationPath>,
+    /// Keystore file containing encrypted private key (default: none).
+    pub keystore: Option<PathBuf>,
+}
+
+/// Proof that a GPG key belongs to the same person as an Ethereum key.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Proof {
+    /// Message to be signed by signee
+    message: String,
+    /// GPG signature of message
+    gpg_sig: String,
+    /// ETH signature of message
+    eth_sig: Signature,
+    /// GPG key fingerprint of the signee
+    gpg_key: String,
+    /// ETH address of the signee
+    eth_key: H160,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    /// No wallet specified.
+    #[error("no wallet specified")]
+    NoWallet,
+    /// Not able to retrieve block .
+    #[error("not able to retrieve block")]
+    NoBlock,
+    /// Not able to retrieve block hash .
+    #[error("not able to retrieve block hash")]
+    NoBlockHash,
+    /// ETH signature failed
+    #[error("eth signature failed")]
+    ETHSigFailed,
+    /// GPG signature failed
+    #[error("{0}")]
+    GPGSigFailed(String),
+}
+
+/// Sign a message with a GPG private key using the GPG CLI
+fn gpg_sign(key: &str, message: &str) -> anyhow::Result<String> {
+    let mut gpg = Command::new("gpg")
+        .arg("--clear-sign")
+        .arg("-u")
+        .arg(key)
+        .stdin(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()?;
+
+    gpg.stdin.as_mut().unwrap().write_all(message.as_bytes())?;
+
+    let output = gpg.wait_with_output()?;
+    if output.status.success() {
+        let raw_output = String::from_utf8_lossy(&output.stdout);
+        Ok(raw_output.to_string())
+    } else {
+        Err(anyhow!(Error::GPGSigFailed(String::from_utf8_lossy(
+            output.stderr.borrow()
+        ).to_string())))
+    }
+}
+
+/// Sign a message with a ETH private key using either a keystore file or a Ledger HW
+pub async fn eth_sign<S: 'static + Signer>(
+    signer: &S,
+    proof: &str,
+) -> Result<Signature, <S as Signer>::Error> {
+    signer.sign_message(proof).await
+}
+
+/// Create the proof to be signed
+fn create_message<T: Display, K: Debug>(
+    ownership: &T,
+    evidence: &K,
+    block_hash: &H256,
+) -> String {
+    format!(
+        "As the owner of GPG key {}, my Ethereum address is {:?} as of {:?}",
+        &ownership, &evidence, &block_hash
+    )
+}
+
+/// Sign the created proof with GPG and ETH keypairs
+async fn sign_proof<T: 'static + Signer>(
+    gpg_key: &str,
+    signer: &T,
+    block_hash: &H256,
+) -> anyhow::Result<Proof> {
+    let message = create_message(&gpg_key, &signer.address(), &block_hash);
+
+    log::info!("Signing message with ETH keypair..");
+    let eth_sig = eth_sign(signer, &message)
+        .await
+        .map_err(|_| anyhow!(Error::ETHSigFailed))?;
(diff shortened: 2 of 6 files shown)


proof-generator/USAGE
```diff
@@ -0,0 +1,20 @@
+Usage
+
+  proof-generator --gpg-key <string> --keystore <file> --rpc-url <url> --output <file> [<option>..] 
+
+Options
+
+    --gpg-key <string>           GPG long key
```
cloudhead, line 7: I thought the key fingerprint was sufficient? Or do we only use it for lookups, but sign the full key?

cloudhead: `gpg -k` ie. the default format seems to be what you expect here..

sebastinez: I changed the argument from `--gpg-fpr` (fingerprint) to `--gpg-key` since we talked about not using the term fingerprint since it could bring confusion.
The idea is that the user still provides the public key fingerprint.

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+/// - Gets either a keystore file or in its absence a Ledger HW as signer to sign a message.
+/// - Creates a message that will be signed by the defined signer.
+/// - Write both proofs to a JSON file.
+pub async fn run(opts: Options) -> anyhow::Result<()> {
+    let provider =
+        Provider::<Http>::try_from(opts.rpc_url).expect("could not instantiate HTTP Provider");
+    let latest_block_number = provider.get_block_number().await?;
+    // 5760 blocks earlier is aprox. 1 day ago, this due to avoid referencing blocks that are affected by reorgs of the chain.
+    let block_number = latest_block_number - 5760u64;
+    if let Some(block) = provider.get_block(block_number).await? {
```
cloudhead, line 139: You can flatten these nested conditionals by using `ok_or` https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// ETH signature failed
+    #[error("eth signature failed")]
+    ETHSigFailed,
+    /// GPG signature failed
+    #[error("{0}")]
+    GPGSigFailed(String),
+}
+
+/// Sign a message with a GPG private key using the GPG CLI
+fn gpg_sign(key: &str, proof: &str) -> anyhow::Result<String> {
```
cloudhead, line 64: The term "proof" seems to be overloaded, here it's simply a *message*, the proof is what you have above under `Proof` plus the keys.

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/// The proof to be constructed and signed
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Proof {
+    /// GPG signature of predefined proof
+    gpg_sig: String,
+    /// ETH signature of predefined proof
+    eth_sig: Signature,
+}
```
cloudhead, line 42: I think verifying the proof will require both public keys, so perhaps they should be included in the proof?

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+/// - Obtains a block hash from a block from 1 day ago.
+/// - Gets either a keystore file or in its absence a Ledger HW as signer to sign a message.
+/// - Creates a message that will be signed by the defined signer.
+/// - Write both proofs to a JSON file.
+pub async fn run(opts: Options) -> anyhow::Result<()> {
+    let provider =
+        Provider::<Http>::try_from(opts.rpc_url).expect("could not instantiate HTTP Provider");
+    let latest_block_number = provider.get_block_number().await?;
+    // 5760 blocks earlier is aprox. 1 day ago, this due to avoid referencing blocks that are affected by reorgs of the chain.
+    let block_number = latest_block_number - 5760u64;
```
cloudhead, line 138: Nitpick: to avoid underflowing when using this against a syncing node that may have less than 5760 blocks, you can use `saturating_sub()`

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+                log::info!("Decrypting keystore..");
+                let prompt = format!("{} Password: ", "??".cyan());
+                let password = rpassword::prompt_password_stdout(&prompt).unwrap();
+                let signer = ethers::signers::LocalWallet::decrypt_keystore(keypath, password)
+                    .map_err(|_| anyhow!("keystore decryption failed"))?;
+                log::debug!("Keystore decrypted: {:?}.", signer);
+
+                let proof = sign_proof(&opts.gpg_key, &signer, &block_hash).await?;
+                fs::write(&opts.output, serde_json::to_string(&proof)?)?;
+                Ok(())
```
cloudhead, line 154: Stylistic thing, but in big blocks of code, I like to keep the last statement separated by a newline to make the code easier to scan, eg.:

```
foo
bar
baz

Ok(())
```

```suggestion

                Ok(())
```

cloudhead (commented):
Good work, just a few suggestions

proof-generator/USAGE
cloudhead: Yeah, I'm referring to the "GPG long key", shouldn't we say "GPG key fingerprint" here?

sebastinez: I understand that the fingerprint is the long key..
We can change it since fingerprint is a more used term for this kind of key.

proof-generator/sign.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -e
+
+exec cargo run -- \
+  --gpg-key EB1729638209DCE61281F416504C9C1DE8C47EDF \
+  --ledger-hdpath "m/44'/60'/0'/0/0" \
+  --output ./proof.json \
+  --rpc-url https://eth-mainnet.alchemyapi.io/v2/yakYiU6gaHXkpqls8Rs9gChnK4Av4Tg8
```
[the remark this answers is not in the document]
sebastinez: Another file I didn't want to commit, I think we can leave it out of the commit

proof-generator/src/lib.rs
sebastinez: Hmm yeah I agree, but I think that showing the GPG public key in the proof would be too much.
We could show the GPG fingerprint which can be used to query a keyserver e.g. to obtain the public key..

proof-generator/src/lib.rs
sebastinez: @cloudhead When using `ok_or` instead of the `if let Some()` I have to return the Error as argument in `ok_or()`
Now since I'm doing the `saturating_sub` and the compiler wants me to use `ethers::types::U64` as integer type. I get the following error
```
the trait bound `Result<_, anyhow::Error>: StdError` is not satisfied
required because of the requirements on the impl of `From<Result<_, anyhow::Error>>` for `anyhow::Error`
required by `std::convert::From::from`rustcE0277
```
Could you give me a little help with that?

cloudhead: Hmm but `saturating_sub` doesn't return a `Result`, it's infallible, maybe that's the problem?

proof-generator/src/lib.rs
cloudhead: Hmmm, yes you would need at least the key fingerprint then, otherwise there's no way to verify this proof without knowing out-of-band what keys signed it. For ethereum you can include the address.
I also think the message that is being signed should be included in here.

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub output: PathBuf,
+    /// RPC url
+    pub rpc_url: String,
+    /// Account derivation path when using a Ledger hardware wallet.
+    pub ledger_hdpath: Option<DerivationPath>,
+    /// Keystore file containing encrypted private key (default: none).
+    pub keystore: Option<PathBuf>,
+}
+
+/// The proof to be constructed and signed
```
cloudhead, line 35:
```suggestion
/// Proof that a GPG key belongs to the same person as an Ethereum key.
```

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    gpg.stdin.as_mut().unwrap().write_all(message.as_bytes())?;
+
+    let output = gpg.wait_with_output()?;
+    if output.status.success() {
+        let raw_output = String::from_utf8(output.stdout)?;
+        Ok(raw_output)
+    } else {
+        Err(anyhow!(Error::GPGSigFailed(String::from_utf8(
+            output.stderr
+        )?)))
```
cloudhead, line 83: When getting the output, I'd use `from_utf8_lossy`, as it's infallible and will just replace unprintable characters with boxes.

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Sign a message with a ETH private key using either a keystore file or a Ledger HW
+pub async fn eth_sign<S: 'static + Signer>(
+    signer: &S,
+    proof: &str,
+) -> Result<Signature, <S as Signer>::Error> {
+    signer.sign_message(proof).await
+}
+
+/// Create the proof to be signed
+fn create_proof<T: Display, K: Debug>(
```
cloudhead, line 96: Again here it's a bit confused, this should be `create_message`, and the below function should be `create_proof`.

You can also just return `String` here.

proof-generator/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+/// - Write both proofs to a JSON file.
+pub async fn run(opts: Options) -> anyhow::Result<()> {
+    let provider =
+        Provider::<Http>::try_from(opts.rpc_url).expect("could not instantiate HTTP Provider");
+    let latest_block_number = provider.get_block_number().await?;
+    // 5760 blocks earlier is aprox. 1 day ago, this due to avoid referencing blocks that are affected by reorgs of the chain.
+    let block_number = latest_block_number.saturating_sub(ethers::prelude::U64::from(5760));
+    let block = provider.get_block(block_number).await?.ok_or(Err(anyhow!(Error::NoBlock)))?; 
+    // TODO: This is redundant eventually, since when a block is found by ethers-rs it must have a hash... I kept it to maintain the code style
+    let block_hash = block.hash.ok_or(Err(anyhow!(Error::NoBlockHash)))?;
```
cloudhead, line 141: Something's not right here, normally `ok_or` would take the `Error` type, not a `Result`, ie. this should be `ok_or(anyhow!(...))`. I wonder if somehow your error is wrapped in another error somehow? :thinking:

proof-generator/src/lib.rs
sebastinez: Okay nevermind, the issue was that I was wrapping the anyhow error with a std error it seems..

proof-generator/src/lib.rs
sebastinez: Yeah this was, the issue I had with `ok_or` now it's working fine

proof-generator/src/lib.rs
cloudhead: You can also change it in the `else` branch

proof-generator/src/lib.rs
sebastinez: Done :+1:

sebastinez:
I'll close this PR and open a new one, since I worked here on master..

PR #6, opened 2021-09-20
Import safe transaction client

sofia-snow:


safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.client
+            .agent
+            .post(&format!(
+                "{}/v1/safes/{}/multisig-transactions/",
+                self.client.transactions_api,
+                to_checksum(&self.safe_address, None),
+            ))
+            .send_json(ureq::json!({
+                "to": to_checksum(&to, None),
+                "value": format!("{}", value),
```
cloudhead, line 152: Isn't this the equivalent of `value.to_string()` or `value.display()`?

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+            }))
+            // 200 OK, is not okay. We expect 201 Created.
+            .and_then(|r| match r.status() {
+                201 => Ok(()),
+                s => Err(ureq::Error::Status(s, r)),
+            })
+            .map_err(From::from)
+    }
+
+    pub fn add_to_confirmations(&self, signed_safe_tx: SignedSafeTxHash) -> Result {
```
cloudhead, line 174: If we're keeping with `propose`, maybe we call this `confirm`? Alternatively I'd rename both to `propose_transaction` and `confirm_transaction` if we want to be explicit.

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+        macro_rules! tokenize {
+            ($($tok:expr),* $(,)*) => {
+                &[$($tok.into_token()),*][..]
+            }
+        }
+
+        let safe_tx_hash = TxHash::from({
+            let mut hasher = Keccak::v256();
+            let mut output = [0u8; 32];
+            hasher.update(&[0x19, 0x01]);
```
cloudhead, line 212: What's this again?

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+            // domain separator
+            hasher.update(&keccak256(&ethers::abi::encode(tokenize![
+                // DOMAIN_SEPARATOR_TYPEHASH
+                U256::from_str(
+                    "0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218"
+                )
```
cloudhead, line 218: Would be nice if these were statically defined at the top of the file, but `from_str` is probably not `const`..

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+    signature.v += 4;
+
+    Ok(SignedSafeTxHash {
+        safe_tx_hash,
+        signature,
+    })
+}
+
+#[test]
+#[cfg(feature = "run-test-against-api")]
```
cloudhead, line 294: I'd prefer to just use `#[ignore]` here

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+fn check() {
+    futures_executor::block_on(async {
```
cloudhead, line 296: This makes me lol

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+        let mut fake_rng =
+            rand_chacha::ChaCha20Rng::from_seed(*b"0123456789abcdef0123456789abcdef");
+        let wallet = LocalWallet::new(&mut fake_rng).with_chain_id(4u64);
+
+        // [x] propose a new transaction
+
+        let value = U256::from(0u64);
+        let _data_unanchor =
+            hex::decode("cbd566010000000000000000000000000000000000000000000000000000000000000000")
+                .unwrap();
```
cloudhead, line 309: Unused var?

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        let safe_tx_hash =
+            TxHash::from_str("0xec812245f4c8908551914922bb1b093b3aa8c02e284c5e8c54fb8170673716cd")
+                .expect("Invalid TxHash");
+
+        let _ = dbg!(safe.add_to_confirmations(
+            sign_tx_hash(&wallet, safe_tx_hash)
+                .await
+                .expect("Invalid signature")
+        ));
```
cloudhead, line 341: Can we remove the `dbg!()`s?

cloudhead:
Looks good overall, just left some suggestions. Going to try the code out asap!

safe-transaction-client/src/lib.rs
sofia-snow: `U256::to_string()` yes, there is no `U256::display`.

safe-transaction-client/src/lib.rs
sofia-snow: The whole crate is about transactions. That'd kinda be like a `Vec::push_element` instead of `Vec::push`. I'll rename it to `confirm`.

safe-transaction-client/src/lib.rs
sofia-snow: The magic bytes? Ask the Gnosis devs.

https://github.com/gnosis/safe-contracts/blob/main/contracts/GnosisSafe.sol#L393

The multi-update hasher is just `keccak256(abi.encodePacked(..))` where the latter function doesn't exist in `ethers::abi` and would just create a vector just to read it back anyway.

safe-transaction-client/src/lib.rs
sofia-snow: `core::convert::FromStr::from_str` is indeed not `const`. However... `U256([u64; 4])` would work, but would not be as trivial to compare against the contract as a copy-paste search. Alternatively the string itself could be the constant.

I prefer the locality for these single-use constants.

Which would you prefer?

safe-transaction-client/src/lib.rs
sofia-snow: Fair. `#[ignore]` or `#[cfg_attr(not(feature = "run-test-against-api"), ignore)]` ?

safe-transaction-client/src/lib.rs
sofia-snow: Sanity check. We can't write `#[test] async fn check() { .. }`. It is certainly not worth depending on `tokio` just for `#[tokio::test]`. :)

safe-transaction-client/src/lib.rs
sofia-snow: Easy to swap between the unanchor and anchor data examples. It is effectively a type-checked comment. I've restructured it for clarity of the choices.

safe-transaction-client/src/lib.rs
sofia-snow: Removed & error handling slightly adjusted for the confirm case -- to actually panic the test if the confirmation failed (but not because the transaction was already executed).

safe-transaction-client/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    pub fn propose(&self, signed_safe_tx: SignedSafeTx) -> Result {
+        let SafeTx {
+            to,
+            value,
+            data,
+            nonce,
+            operation,
+            ..
+        } = signed_safe_tx.inner;
```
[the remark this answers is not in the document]
sofia-snow: Done.

safe-transaction-client/src/lib.rs
cloudhead: Yeah kind of weird there isn't a `futures::test` decorator ...

safe-transaction-client/src/lib.rs
cloudhead: Ah it's `eth_sign`: https://eips.ethereum.org/EIPS/eip-191

safe-transaction-client/src/lib.rs
cloudhead: yeah... let's keep it like this I guess.

PR #7, opened 2021-09-21
Anchor with Safe

cloudhead:


safe-transaction-client/src/lib.rs
```diff
@@ -89,11 +100,42 @@ impl Client<'_> {
...
+                "{}/v1/safes/{}/",
+                self.transactions_api,
+                to_checksum(&safe_address, None),
+            ))
+            .call()?
+            .into_json()?;
+
+        let owners = owners
+            .iter()
+            .map(|o| Address::from_str(o))
```
sofia-snow, line 127: You may use `.map(Address::from_str)`.

safe-transaction-client/src/lib.rs
```diff
@@ -105,28 +147,12 @@ impl Safe<'_> {
         data: Bytes,
         operation: Operation,
     ) -> Result<SafeTx> {
```
sofia-snow, line 149: `Result` is now unnecessary.

safe-transaction-client/src/lib.rs
cloudhead: Actually this is a case where you can't for some annoying reason:

```rust
error[E0631]: type mismatch in function arguments
   --> safe-transaction-client/src/lib.rs:115:18
    |
115 |             .map(Address::from_str)
    |                  ^^^^^^^^^^^^^^^^^
    |                  |
    |                  expected signature of `fn(&std::string::String) -> _`
    |                  found signature of `for<'r> fn(&'r str) -> _`
```

Issue #8, opened 2021-09-23
NFT Reward Contribution Mechanism


sebastinez:
## Abstract
This issue details the implementation of a proof of contribution mechanism, that would allow a user to claim a NFT showing his contribution through a commit to a specific organization.

## Glossary
- Message = The message describing the contribution.
- Proof = The hashed and by the Orgs member(s) signed message.
- NFT factory = The smart contract responsible to mint and transfer the contribution rewards.

## Discovery
- A org MAY, using the CLI search their repos for signed commits, in the commit header and in the trailers of the commit message.
    ```
    $ rad-reward discover --path <path-to-repo>
    Commits                                    Contributions
    f673cf386bc01688c53b51b4c2eedcf2945d70ea > committed-by  0CE7AA44C56AE6A8C43978663E3C1B6185AC4475
    ef27096c91f95791dccaf9992ef772b28a76558e > committed-by  0x8152237402E0f194176154c3a6eA1eB99b611482
                                               authored-by   0x5E813e48a81977c6Fdd565ed5097eb600C73C4f0
    f179aab522d746735d73a8342a12a276be0e6cb8 > committed-by  ABAF11C65A2970B130ABE3C479BE3E4300411886
    ```
- The CLI MUST show a listing of all the eligible commits, and all the signed contributions that were made.

## Creation
- With a commit found through the discovery mechanism, an org MAY create an offchain proof.
  ```
  $ rad-reward create \
    --receiver <receiver-address> \
    --org <org-address> 
    --project-urn <project-urn> \
    --commit <commit-hash> // provided or selected through CLI
  ```
- The message to be signed MUST follow this pattern: 
    ```
    The org {org_address} wants to reward {receiver_address} for the contribution of commit {commit_hash} in the project {project_id}
    ```

    >**Example:**
    > 
    > The org 0x8152237402E0f194176154c3a6eA1eB99b611482 wants to reward 0x5E813e48a81977c6Fdd565ed5097eb600C73C4f0 for the contribution of commit b90bf64bafb655d5e8b30fcd156b014c3922921b in the project rad:git:hnrkybhzyqzg37cmaoyh7qs8uayjzf4qn7w1y
- The org MUST hash the message with SHA256, sign the resulting hash by all its members and provide the resulting signatures:
  - In the case of a single owner, the owner MUST sign the hashed data.
  - In the case of a gnosis safe each member of the safe MUST sign the hashed data.

  *Example of a proof for a single owner org*
  ```json
[...]

PR #12, opened 2021-10-18
Code Contribution Reward CLI

sebastinez:
This PR works on #8 

It adds the reward CLI that allows in its present state:
- The creation by an org of a cryptographic puzzle representing the contribution of a specific user to a project.
- Claiming as a contributor a stored proof on a repo and minting the NFT corresponding to this contribution.

Open issues:
- [x] Refactor the `claim_with_ledger` and `claim_with_keystore` into a sole function and define the type used in the claim function. Thanks @Ryanmtate!
- [x] Refactor parts of the tx handling to avoid reusing code  Thanks @Ryanmtate!

**On the tokenURI part:**
  We could eventually provide a option to only provide the `id` e.g.: the CID, git URL, etc and the CLI could parse it and identify it, and so create the corresponding tokenURI

This PR will implement the ability to  claim a NFT as EOA.
In the next iterations we'll allow Safes and other Contracts to claim contribution rewards for their work, and in a near future eventually the automation of creating and claiming, using the anchor functionality.

Open to feedback


Cargo.toml
```diff
@@ -26,10 +26,6 @@ branch = "master"
 git = "https://github.com/radicle-dev/radicle-link"
 branch = "master"
 
-[patch.crates-io.ethers]
-git = "https://github.com/gakonst/ethers-rs"
-branch = "master"
-
```
sebastinez, line 31: To allow using a git dependency in the `reward` CLI, the only way to solve it was to remove this patch.
Will see if I find a better way to make this work since this eventually brakes other CLIs in this repo..

sebastinez:
@Ryanmtate Please feel free to review the PR and let me know what you think.

Cargo.toml
sebastinez: After talking with @Ryanmtate I think we should avoid patching ethers in the global Cargo.toml and let each tool define ethers for itself.

Ryanmtate: Eventually good to standardize the client-tools on a single version of ethers-rs, otherwise will have conflicting versions as a dependency.

Maybe instead we do include the patch, but on the `rev` you need.

sebastinez: never mind I was able to fix the conflicts I had with the latest commits on the ethers master branch so I brought the patch back

reward/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "reward"
```
cloudhead, line 2:
```suggestion
name = "rad-reward"
```

reward/USAGE
```diff
@@ -0,0 +1,43 @@
+Usage
+
+  rad-reward [<option>..] [subcommand] 
+
+Commands
+
+    create                       Create a new self contained proof and attach it to a commit
```
cloudhead, line 7: Should this be `--create`? Seems like it based on the example

reward/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+#[derive(Debug, Clone)]
+pub struct Options {
+    /// Address of org.
+    pub org: Option<Address>,
+    /// Address of contributor
+    pub contributor: Option<Address>,
+    /// Repo path
+    pub repo: Option<PathBuf>,
+    /// Project id.
+    pub project: Option<String>,
```
cloudhead, line 360: Are `repo` and `project` ever optional? How about `org`?

cloudhead (commented):
Looks good! Few questions/comments.

reward/src/lib.rs
sebastinez: I left everything optional, since `--create` needs almost all the arguments, `--claim` needs only `--repo`.

cloudhead (approved):
Looks good, just needs build fixes.

sebastinez:
There are some CI issues but not caused by the reward CLI.
I would like to tackle these in a separate issue

cloudhead:
Yeah, this is basicall V2 of `rad reward`, so will revisit at some point in the future.

PR #15, opened 2021-11-04
remote-helper: Support project IDs

cloudhead:


remote-helper/src/main.rs
```diff
@@ -3,24 +3,29 @@ use ethers::contract::abigen;
 use ethers::prelude::*;
 use ethers::types::{Address, NameOrAddress};
 
+use librad::crypto::{BoxedSigner, SomeSigner};
 use link_identities::git::Urn;
+use radicle_git_helpers::remote_helper;
+
+use rad_remote_helper::signer::Signer;
```
Ryanmtate, line 10: Is this the correct signer? It seems like we are importing this crate itself? Unless I am mistaken.

PR #17, opened 2021-12-09
Implements CLI protoype

erak:
### Description

This is a CLI prototype build with Python and [click](https://github.com/pallets/click). It assumes that you have the build folders, containing radicle (development) binaries added to your PATH. It relies on the radicle plumbing binaries returning proper JSON output (still WIP).

### Discussion

- We should probably use `cargo run --quiet --bin <bin>` instead of calling the pre-built binaries directly (needed during development)
- How to install remote helper: `git-remote-rad`


rad-prototype/rad.py
```diff
@@ -0,0 +1,111 @@
...
+import json
+
+import click
+from click.decorators import password_option
+
+import ui, render
+
+RAD_HOME = os.getenv('RAD_HOME')
+
+def valid_env():
```
erak, line 16: `valid_env` and `profile_exists` should be moved into separate file.

cloudhead:
This has been implemented in rust, closing.

PR #18, opened 2022-01-07
[CLI] rad-auth, rad-init and rad-publish

erak:
### Description

This PR implements the first set of binaries needed for the non-P2P onboarding flow:

- `rad-auth`: Creates new profile and identity (and monorepo if it does not exist).
- `rad-init`: Initializes a new project in the monorepo when called inside an existing git repository.
- `rad-publish`: Pushes current project from monorepo to given seed (based on URN from Radicle entry in git config, which is created by `rad-init`).

#### Terminal UI
For the implementation of the terminal UI (program output and user input), the crates `dialoguer` and `indicatif`, both maintained by the same people, were used. They are based on the terminal abstraction crate `console` and provide a high-level interface. 

Both were chosen in favor of a combination of `crossterm-rs` (too low level) and `spinners`.

### Building
**You will run into an compilation error in `rad-clib` together with `thrussh`. In order to fix this, you need to checkout [`feature/rad-cli-deps-thrussh`](https://github.com/erak/radicle-client-tools/tree/feature/cli-v1), which contains the temporary switches to my fork.**

#### General
- [ ]  Compile README with detailed usage
- [ ] Document library code
- [ ] Clarify `thrussh` bugfix situation
- [ ] Split-up `lib.rs` into single files

#### `rad-auth`
- [ ] Provide information about required git config (key, ssh signing, etc.)
- [ ] Clean up error handling

#### `rad-init`
- [ ] Look into `id` file implementation to signal radicle project

#### `rad-publish`
- [ ] Implement seed selection

#### `rad-sync`
- [ ] Once a (project, peer) is tracked, rad sync should fetch that remote peer for the given project


.gitignore
```diff
@@ -1,2 +1,4 @@
 /target
+*/target
 errors.err
+Cargo.lock
```
cloudhead, line 4: It's useful to check Cargo.lock in to have reliable CI and stuff.

rad-cli/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "rad-auth"
+path = "src/bin/auth.rs"
+
+[[bin]]
+name = "rad-init"
+path = "src/bin/init.rs"
+
+[[bin]]
+name = "rad-sync"
+path = "src/bin/sync.rs"
```
cloudhead, line 28: Newline

rad-cli/src/bin/auth.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use rad_clib::keys::ssh::SshAuthSock;
+
+use rad_cli::{keys, person, proc::some_or_exit, profile, tui};
+
+#[derive(Debug, StructOpt)]
+pub struct Args {
+    #[structopt(short, long)]
+    pub add: bool,
+}
```
cloudhead, line 18: I'm ok with using structopt for now as long as we don't invest too heavily in it, but eventually I think I'd like all CLIs to use `lexopt` which we use in some of the other crates. The reason is that declarative configuration often isn't flexible enough to encode all the logic and you end up having to have a second phase of argument processing after the initial one.

rad-cli/src/bin/sync.rs
```diff
@@ -0,0 +1,53 @@
...
+use rad_cli::{proc::some_or_exit, keys, id, tui, profile, project, seed};
+
+fn main() -> anyhow::Result<()> {
+    tui::headline("Syncing your local 🌱 projects");
+
+    let seed = "http://localhost:8778".to_string();
+    let home = RadHome::default();
+
+    let profile = some_or_exit(profile::default());
+    let storage = some_or_exit(keys::storage(&profile, SshAuthSock::default()));
```
cloudhead, line 16: I have to say, I'm not a big fan of this pattern. Rust has the `?` syntax for this use-case. `Option` can also be converted to `Result` if needed, via `ok_or`.

rad-cli/src/bin/init.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright © 2021 The Radicle Link Contributors
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
```
cloudhead, line 4: License header should be removed.

rad-cli/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+        pub fn error_detail(detail: &str) {
+            println!("  {} {}", style("⊙").red(), &detail);
+        }
+    }
+}
+
+pub mod id {
+    use librad::git::Urn;
+    pub fn from_urn(urn: &Urn) -> String {
+        urn.to_string().split(":").last().unwrap().to_string()
```
cloudhead, line 433: `Urn::encode_id` is what you want I think.

.gitignore
erak: Ah, that was by mistake. Thanks.

rad-cli/Cargo.toml
erak: This file will go away.

rad-cli/src/bin/init.rs
erak: Done.

rad-cli/src/bin/sync.rs
erak: Ok. Agree. Found the right way to report through the CLI UI.

rad-cli/src/lib.rs
erak: Great! Thanks.

rad-cli/src/bin/auth.rs
erak: Ok, thanks for the explanation. I switched to `lexopt` already.

rad-auth/src/args.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    let mut add = false;
+    let mut parser = lexopt::Parser::from_env();
+    while let Some(arg) = parser.next()? {
+        match arg {
+            Long("add") => {
+                add = true;
+            }
+            Long("help") => {
+                println!("Usage: hello [-n|--number=NUM] [--shout] THING");
```
erak, line 17: Needs to be fixed.

rad-auth/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+use cli::{keys, person, profile, tui};
+
+mod args;
+
+fn main() -> anyhow::Result<()> {
+    match run() {
+        Ok(()) => Ok(()),
+        Err(_) => {
+            println!();
+            std::process::exit(0);
```
cloudhead, line 12:
```suggestion
            std::process::exit(1);
```
Non-zero exit for errors.

Cargo.toml
```diff
@@ -4,9 +4,20 @@ version = "0.0.0"
...
+git2 = "0.13.20"
 colored = { version = "1.9" }
 atty = { version = "0.2" }
 log = { version = "0.4", features = ["std"] }
+dialoguer = "0.9.0"
+indicatif = "0.16.2"
+librad = { path = "../radicle-link/librad" }
+rad-clib = { path = "../radicle-link/rad-clib" }
+rad-profile = { path = "../radicle-link/rad-profile" }
+rad-identities = { path = "../radicle-link/rad-identities" }
```
cloudhead, line 17: Dependencies for the radicle libs for now should be specified with `version = "0"`, and a patch should be used in this file. Happy to explain more in RTC.

rad-auth/src/main.rs
erak: :facepalm: That sneaked in...

cloudhead:
Replaced by #21 #22 and #23

PR #33, opened 2022-02-04
Include the man page in the .deb package

adaszko:


.github/workflows/package.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - adaszko/*
```
cloudhead, line 7:
```suggestion
```

cloudhead (dismissed):

PR #35, opened 2022-02-08
rad-ci: Initial commit

xphoniex:
Needs some more work, but ready to be reviewed.


src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+        let identity = identities.get(options.ssh_index).context(format!(
+            "Couldn't retrieve ssh-agent key #{}",
+            options.ssh_index
+        ))?;
+        agent.userauth(&options.ssh_user, identity)?;
+    }
+    if sess.authenticated() {
+        spinner.finish();
+    } else {
+        spinner.failed();
```
cloudhead, line 41: No point in continuing if we couldn't auth.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+use rad_terminal::components as term;
+
+use anyhow::Context;
+use ssh2::Session;
+
+fn main() {
+    rad_terminal::args::run_command::<Options, _>(HELP, "CI", run);
+}
+
+fn run(options: Options) -> anyhow::Result<()> {
```
cloudhead, line 15: This should be moved to the `lib.rs`, along with any dependencies

help/src/lib.rs
```diff
@@ -23,6 +23,7 @@ const COMMANDS: &[Help] = &[
     rad_untrack::HELP,
     rad_sync::HELP,
     rad_ens::HELP,
+    rad_ci::HELP,
```
cloudhead, line 26: Let's not include it in the list of core commands yet, until it gets more testing.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+        let mut agent = sess.agent()?;
+        agent.connect()?;
+        agent.list_identities()?;
+        let identities = agent.identities()?;
+        let identity = identities.get(options.ssh_index).context(format!(
+            "Couldn't retrieve ssh-agent key #{}",
+            options.ssh_index
+        ))?;
+        agent.userauth(&options.ssh_user, identity)?;
```
cloudhead, line 36: Passing a key index is kind of weird, can we not pass a key path?

cloudhead (commented):
This is cool, left some comments and will try it out when I have time.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+                None,
+            )
+            .unwrap();
+
+        remote_file.write_all(content.as_bytes()).unwrap();
+
+        remote_file.send_eof().unwrap();
+        remote_file.wait_eof().unwrap();
+        remote_file.close().unwrap();
+        remote_file.wait_close().unwrap();
```
cloudhead, line 98: You can get rid of the unwraps here and in other places.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+fn run(options: Options) -> anyhow::Result<()> {
+    let seed = options.seed.context("Seed is invalid")?;
+    // SSH into server
+    let tcp = TcpStream::connect(format!("{}:22", &seed))?;
+    let mut sess = Session::new()?;
+    sess.set_tcp_stream(tcp);
+    sess.handshake()?;
+
+    // Authenticate
+    let spinner = term::spinner(&format!("SSH into remote {}", &seed));
```
cloudhead, line 24: Spinner should probably start before the TCP stream is connected.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        term::info!("Requirements are not installed.");
+        std::process::exit(1);
+    }
+    term::blank();
+
+    // Upload file if supplied
+    if options.config.is_some() {
+        let path = options.config.unwrap();
+        let mut file = std::fs::File::open(path).context("Couldn't open config file")?;
```
cloudhead, line 79: We usually import stuff at the top level to make code more readable, same with `Path`

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+        remote_file.wait_eof().unwrap();
+        remote_file.close().unwrap();
+        remote_file.wait_close().unwrap();
+    }
+
+    // Apply changes
+    if options.install {
+        // Copy the docker-compose.yml over
+        let spinner = term::spinner("Downloading docker-compose.yml for Concourse CI");
+        execute_fn("wget -nc -O /etc/concourse-docker-compose.yml https://concourse-ci.org/docker-compose.yml", spinner);
```
cloudhead, line 105: Move URL to top-level constant

ci/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut install = false;
+        let mut uninstall = false;
+        let mut seed = None;
+        let mut ssh_user = String::from("root");
```
cloudhead, line 45: It's safer and less surprising to login as the current user and use `sudo` if priviledges are needed.

src/bin/rad-ci.rs
xphoniex: It doesn't reach the `else`, either gets:

```
== CI failed [Session(-34)] no identities found in the ssh agent
```
or
```
== CI failed [Session(-18)] Username/PublicKey combination invalid
```

src/bin/rad-ci.rs
xphoniex: https://github.com/radicle-dev/radicle-client-tools/pull/35/files#diff-2f5300f7dedf306f4b73a94a6c5c3674f3ece69ed4bda39a3b27b8448149d978R119

help/src/lib.rs
xphoniex: done

src/bin/rad-ci.rs
xphoniex: (I think avoiding key path in favor of `ssh-agent` is a better option, now that I used it)

src/bin/rad-ci.rs
xphoniex: like this? https://github.com/radicle-dev/radicle-client-tools/pull/35/commits/21aa4f338dfbf8b7570287a4f7701d4462f48278#diff-2f5300f7dedf306f4b73a94a6c5c3674f3ece69ed4bda39a3b27b8448149d978R201-R204

src/bin/rad-ci.rs
xphoniex: yup https://github.com/radicle-dev/radicle-client-tools/pull/35/commits/21aa4f338dfbf8b7570287a4f7701d4462f48278#diff-2f5300f7dedf306f4b73a94a6c5c3674f3ece69ed4bda39a3b27b8448149d978R122

src/bin/rad-ci.rs
xphoniex: done

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,161 @@
...
+    if options.config.is_some() {
+        let path = options.config.unwrap();
+        let mut file = std::fs::File::open(path).context("Couldn't open config file")?;
+        let mut content = String::new();
+        file.read_to_string(&mut content)
+            .context("Couldn't read the contents of config file")?;
+
+        let mut remote_file = sess
+            .scp_send(
+                std::path::Path::new("/etc/concourse-docker-compose.yml"),
```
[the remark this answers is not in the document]
xphoniex: actually, I think we should use `/app/radicle/etc`, what do you think?

src/bin/rad-ci.rs
xphoniex: done

ci/src/lib.rs
xphoniex: on initial spinner, I now log which user we're ssh'ing using, so if it's the wrong user they can change it https://github.com/radicle-dev/radicle-client-tools/pull/35/commits/21aa4f338dfbf8b7570287a4f7701d4462f48278#diff-2f5300f7dedf306f4b73a94a6c5c3674f3ece69ed4bda39a3b27b8448149d978R122-R126

also added `whoami` https://github.com/radicle-dev/radicle-client-tools/pull/35/commits/21aa4f338dfbf8b7570287a4f7701d4462f48278#diff-2f5300f7dedf306f4b73a94a6c5c3674f3ece69ed4bda39a3b27b8448149d978R58

xphoniex:
[`receive-hook`](https://github.com/radicle-dev/radicle-client-tools/blob/e1c97710b13020c5a55a1ba8aa40b66caa1894f4/ci/receive-hook#L6) needs an .env file to source user/pass and url of concourse server so I added [`extract-env`](https://github.com/radicle-dev/radicle-client-tools/blob/e1c97710b13020c5a55a1ba8aa40b66caa1894f4/ci/extract-env) which uses `grep` and `cut` to extract those values from yaml file. Not ideal, but works for now.

src/bin/rad-ci.rs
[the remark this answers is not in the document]
cloudhead: Also problematic, it won't work unless they have a docker install with our docker images.

src/bin/rad-ci.rs
```diff
@@ -0,0 +1,5 @@
+use rad_ci::{run, Options, HELP};
+
+fn main() {
+    rad_terminal::args::run_command::<Options, _>(HELP, "ci", run);
```
cloudhead, line 4: The string here is prepended to failure messages like this: `{string} failed: {error}`.

So let's set it to just "Command" if this tool can do multiple things, then it'll show `Command failed: {error}`.

src/bin/rad-ci.rs
cloudhead: Heh no, by using the try operator: `remote_file.send_eof()?` etc.

ci/receive-hook
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+echo "**** ***** post-receive hook ***** ****"
```
cloudhead, line 3: Maybe `echo "Running radicle CI hook..."` would be more informative

ci/receive-hook
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
```
cloudhead, line 1: To maximize system compatibility I'd recommend only depending on `#!/bin/sh` (lots of people have `dash` instead of bash).

Also maybe makes sense to `set -ue` to fail fast if there's anything unexpected.

ci/receive-hook
```diff
@@ -0,0 +1,104 @@
...
+if [ -f "/app/radicle/.ci.env" ]; then
+	source /app/radicle/.ci.env
+fi
+
+run_pipeline() {
+  local urn="$RADICLE_NAME-${GIT_NAMESPACE:30}"
+  # login
+  fly -t local login -c $CONCOURSE_URL -u $CONCOURSE_USER -p $CONCOURSE_PASS
+  # create team for project
+  #fly -t local set-team -n $urn --local-user $CONCOURSE_USER --non-interactive
```
cloudhead, line 14: Can we get rid of the commented code in this file?

ci/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        // Copy receive-hook file over
+        let spinner = term::spinner("Copying receive-hook file to server");
+        let receive_hook_content = include_str!("../receive-hook");
+        upload_content_to_file(
+            &sess,
+            receive_hook_content.as_bytes(),
+            "/app/radicle/hooks/receieve-hook",
+            0o755,
+        );
```
cloudhead, line 233: This is not going to work, as you don't know how `--root` was configured on the server. I think it's best to either have a `--remote-root` flag to specify it, or to let users install this file manually.

src/bin/rad-ci.rs
xphoniex: Are you sure about this? won't be too informative, also as of now, it's only doing install/uninstalls.

src/bin/rad-ci.rs
xphoniex: I went ahead and removed index, instead now we iterate over all keys.

xphoniex:
added `--root` argument which should fix issues with paths, etc.

PR #39, opened 2022-02-11
Add Homebrew formula

adaszko:


cloudhead (commented):
Nice. Can we put this in `/pkg/brew/radicle-tools.rb`?

HomebrewFormula/radicle-tools.rb
```diff
@@ -0,0 +1,31 @@
+class RadicleTools < Formula
+  version "0.0.0"
+  desc "Radicle CLI tools"
+  homepage "https://github.com/radicle-dev/radicle-client-tools"
```
cloudhead, line 4: Let's do `https://app.radicle.network/alt-clients.radicle.eth/radicle-client-tools`

HomebrewFormula/radicle-tools.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  url "https://github.com/radicle-dev/radicle-client-tools/releases/download/v#{version}/radicle-tools-x86_64-apple-darwin.tar.gz"
```
cloudhead, line 6: We should start to think about how we could host binaries on radicle :thinking:

adaszko:
> Nice. Can we put this in `/pkg/brew/radicle-tools.rb`?

Done ✅

HomebrewFormula/radicle-tools.rb
adaszko: Maybe.  That sounds like a task to be considered in its own right though, so maybe let's put it on a backlog?

cloudhead: Aye, just thinking aloud

PR #49, opened 2022-02-22
Update man pages

cloudhead:
Signed-off-by: Alexis Sellier <self@cloudhead.io>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,9 @@ assets = [
     ["target/release/rad-untrack", "usr/bin/rad-untrack", "755"],
     ["target/release/rad-remote", "usr/bin/rad-remote", "755"],
     ["target/release/git-remote-rad", "usr/bin/git-remote-rad", "755"],
-    ["radicle-tools.1.gz", "usr/share/man/man1/radicle-tools.1.gz", "644"],
+    ["rad.1.gz", "usr/share/man/man1/rad.1.gz", "644"],
+    ["rad-checkout.1.gz", "usr/share/man/man1/rad-checkout.1.gz", "644"],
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -24,7 +24,8 @@ assets = [
     ["target/release/rad-untrack", "usr/bin/rad-untrack", "755"],
     ["target/release/rad-remote", "usr/bin/rad-remote", "755"],
     ["target/release/git-remote-rad", "usr/bin/git-remote-rad", "755"],
-    ["radicle-tools.1.gz", "usr/share/man/man1/radicle-tools.1.gz", "644"],
+    ["rad.1.gz", "usr/share/man/man1/rad.1.gz", "644"],
+    ["checkout/rad-checkout.1.gz", "usr/share/man/man1/rad-checkout.1.gz", "644"]
```
adaszko, line 28: The compiled man pages are located in the top-level directory:
```suggestion
    ["rad-checkout.1.gz", "usr/share/man/man1/rad-checkout.1.gz", "644"]
```

pkg/brew/radicle-tools.rb
```diff
@@ -22,7 +21,8 @@ def install
     bin.install "rad-checkout"
     bin.install "rad-untrack"
     bin.install "git-remote-rad"
-    man1.install "radicle-tools.1.gz"
+    man1.install "rad.1.gz"
+    man1.install "rad-checkout.1.gz"
```
adaszko, line 25: We have a chicken and egg problem here.  Merging this hunk is going to break current `brew install ...` because the `url` above points to a tarball that doesn't have those files yet.  OTOH, the tarball having those files doesn't exist yet because it's produced when a git tag is pushed.  So: This hunk would have to land in the commit tagged `vx.x.x`.

PR #58, opened 2022-03-17
Add support for WalletConnect in `rad ens`

xphoniex:
Currently in `Cargo.toml` it's using:

```
walletconnect = { git = "https://github.com/xphoniex/walletconnect-rs", branch = "v0.1.0", features = ["qr"] }`. 
```

which is in my own personal github. Might be helpful to move this to `radicle-dev`, before merging. I changed the types in original package and since author is not responsive, not sure we can merge it in.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -6,7 +6,7 @@ license = "GPL-3.0-or-later"
 
 [features]
 default = []
-ethereum = ["ethers", "coins-bip32", "async-trait", "hex"]
+ethereum = ["ethers", "coins-bip32", "async-trait", "hex", "walletconnect"]
 
 [dependencies]
 anyhow = "1.0"
@@ -48,4 +48,10 @@ optional = true
 
 [dependencies.hex]
 version = "0.4.3"
+optional = true
+
+[dependencies.walletconnect]
+git = "https://github.com/xphoniex/walletconnect-rs"
+branch = "v0.1.0"
+features = ["qr"]
 optional = true
--- a/common/src/ethereum.rs
+++ b/common/src/ethereum.rs
@@ -18,12 +18,17 @@ use ethers::types::Chain;
 use rad_terminal::args;
 use rad_terminal::components as term;
 
+mod walletconnect;
+
+use self::walletconnect::WalletConnect;
+
 /// Radicle's ENS domain.
 pub const RADICLE_DOMAIN: &str = ".radicle.eth";
 
 pub const SIGNER_OPTIONS: &str = r#"
     --ledger-hdpath <hdpath>     Account derivation path when using a Ledger hardware device
     --keystore <file>            Keystore file containing encrypted private key (default: none)
+    --walletconnect              Use WalletConnect
 "#;
 
 pub const PROVIDER_OPTIONS: &str = r#"
@@ -42,6 +47,8 @@ pub struct SignerOptions {
     pub ledger_hdpath: Option<DerivationPath>,
     /// Keystore file containing encrypted private key (default: none).
     pub keystore: Option<PathBuf>,
+    /// Walletconnect account (default: false).
+    pub walletconnect: bool,
 }
 
 impl SignerOptions {
@@ -54,6 +61,7 @@ impl SignerOptions {
             ledger_hdpath: env::var("ETH_HDPATH")
                 .ok()
                 .and_then(|v| DerivationPath::from_str(v.as_str()).ok()),
+            walletconnect: false,
         };
 
         while let Some(arg) = parser.next()? {
@@ -69,6 +77,9 @@ impl SignerOptions {
                     let value = parser.value()?;
 
                     options.keystore = Some(args::parse_value(&flag, value)?);
+                }
+                Long("walletconnect") => {
+                    options.walletconnect = true;
                 }
                 _ => unparsed.push(args::format(arg)),
             }
@@ -130,6 +141,8 @@ pub enum WalletError {
     Ledger(#[from] LedgerError),
     #[error(transparent)]
     Local(#[from] ethers::signers::WalletError),
+    #[error(transparent)]
+    WalletConnect(#[from] ::walletconnect::client::CallError),
     #[error("no wallet specified")]
     NoWallet,
 }
@@ -139,6 +152,7 @@ pub enum WalletError {
 pub enum Wallet {
     Ledger(Ledger),
     Local(LocalWallet),
+    WalletConnect(WalletConnect),
 }
 
 #[async_trait::async_trait]
@@ -149,6 +163,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => s.chain_id(),
             Self::Local(s) => s.chain_id(),
+            Self::WalletConnect(s) => s.chain_id(),
         }
     }
 
@@ -156,6 +171,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => s.address(),
             Self::Local(s) => s.address(),
+            Self::WalletConnect(s) => s.address(),
         }
     }
 
@@ -163,6 +179,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => Self::Ledger(s.with_chain_id(chain_id)),
             Self::Local(s) => Self::Local(s.with_chain_id(chain_id)),
+            Self::WalletConnect(_s) => unimplemented!(),
         }
     }
 
@@ -173,6 +190,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => s.sign_typed_data(payload).await.map_err(WalletError::from),
             Self::Local(s) => s.sign_typed_data(payload).await.map_err(WalletError::from),
+            Self::WalletConnect(_s) => unimplemented!(),
         }
     }
 
@@ -183,6 +201,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => s.sign_message(message).await.map_err(WalletError::from),
             Self::Local(s) => s.sign_message(message).await.map_err(WalletError::from),
+            Self::WalletConnect(s) => s.sign_message(message).await.map_err(WalletError::from),
         }
     }
 
@@ -193,6 +212,7 @@ impl Signer for Wallet {
         match self {
             Self::Ledger(s) => s.sign_transaction(message).await.map_err(WalletError::from),
             Self::Local(s) => s.sign_transaction(message).await.map_err(WalletError::from),
+            Self::WalletConnect(s) => s.sign_transaction(message).await.map_err(WalletError::from),
         }
     }
 }
@@ -223,6 +243,13 @@ impl Wallet {
                 .context("Could not connect to Ledger device")?;
 
             Ok(Wallet::Ledger(signer))
+        } else if options.walletconnect {
+            let signer = WalletConnect::new()
+                .map_err(|_| anyhow!("Failed to create WalletConnect client"))?
+                .show_qr()
+                .await
+                .context("Failed to connect to WalletConnect session")?;
+            Ok(Wallet::WalletConnect(signer))
         } else {
             Err(WalletError::NoWallet.into())
         }
--- a/common/src/ethereum/walletconnect.rs
+++ b/common/src/ethereum/walletconnect.rs
@@ -1,0 +1,104 @@
+use std::error::Error;
+use walletconnect::client::{CallError, SessionError};
+use walletconnect::{qr, Client, Metadata, Transaction};
+
+use ethers::types::transaction::eip2718::TypedTransaction;
+use ethers::types::{Address, NameOrAddress, Signature, U256};
+
+#[derive(Debug)]
+pub struct WalletConnect {
+    client: Client,
+    chain_id: u64,
+    address: Address,
+}
+
+impl WalletConnect {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        let client = Client::new(
+            "radicle-cli",
+            Metadata {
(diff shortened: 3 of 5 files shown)


common/Cargo.toml
```diff
@@ -28,6 +28,7 @@ url = { version = "2" }
 sha2 = { version = "0.10.2" }
 ureq = { version = "2.2", default-features = false, features = ["json", "tls"] }
 thiserror = "1"
+walletconnect = { git = "https://github.com/xphoniex/walletconnect-rs", branch = "v0.1.0", features = ["qr"] }
```
cloudhead, line 31: Let's move this below and make it optional like the other ethereum-only deps.

common/src/ethereum.rs
```diff
@@ -132,13 +141,16 @@ pub enum WalletError {
     Local(#[from] ethers::signers::WalletError),
     #[error("no wallet specified")]
     NoWallet,
+    #[error(transparent)]
+    CallError(#[from] walletconnect::client::CallError),
```
cloudhead, line 145: Move above `NoWallet` and call it `WalletConnect` to match the other variants

common/src/ethereum.rs
```diff
@@ -236,7 +262,7 @@ where
     M: Middleware + 'static,
 {
     let receipt = loop {
-        let spinner = term::spinner("Waiting for transaction to be signed...");
+        let spinner = term::spinner("Waiting for transaction to be signed & sent...");
```
cloudhead, line 265: I would revert this, we're not waiting for it to be sent, since the sending is instant once it is signed, and once it's signed we log that it's been sent.

cloudhead:
Nice. Can you link to the PR on walletconnect? We can move it to radicle-dev and our seed.

common/Cargo.toml
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/6bd4b556fdbdb58f9af097f15e2af011db40ed5d#diff-a51b8c3d64e9beafa9b726f1fc1c9099201507aef512f67724a1ac531573d4afR9-R57

common/src/ethereum.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/6bd4b556fdbdb58f9af097f15e2af011db40ed5d#diff-799f6adcb3a8a0469e7950f80331d4ffcb715140d420d1f4b25bad88ee61adc9R144-R145

common/src/lib.rs
```diff
@@ -5,6 +5,7 @@ pub mod person;
 pub mod profile;
 pub mod project;
 pub mod seed;
+pub mod walletconnect;
```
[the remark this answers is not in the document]
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/6bd4b556fdbdb58f9af097f15e2af011db40ed5d#diff-799f6adcb3a8a0469e7950f80331d4ffcb715140d420d1f4b25bad88ee61adc9R21

common/src/ethereum.rs
xphoniex: The reason I changed this is because when using `WalletConnect` there's a delay between when you sign tx in walletconnect and when it's sent back to app to be sent to blockchain. When there was an error with tx info, e.g. gas being too low, it'd still say "Waiting for transaction to be signed..." failed which is not very helpful, that's why I added "... signed & sent". Makes sense?

xphoniex:
> Nice. Can you link to the PR on walletconnect? We can move it to radicle-dev and our seed.

Here's the PR https://github.com/nlordell/walletconnect-rs/pull/3 .

Also tried to create a PR on `ethers` but that too is idle as well. 

You can fork [this](https://github.com/xphoniex/walletconnect-rs/tree/v0.1.0) and then we can use its url.

common/src/ethereum.rs
cloudhead: Ah I see, that's interesting, if there's no way to wait on the intermediate step, then I'd just say `Waiting for transaction to be submitted`

ens/src/lib.rs
```diff
@@ -320,7 +322,7 @@ async fn setup(
         }
     }
 
-    let call = resolver.multicall(calls)?;
+    let call = resolver.multicall(calls)?.gas(21000);
```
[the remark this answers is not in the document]
cloudhead: Weird.

common/src/ethereum/walletconnect.rs
```diff
@@ -0,0 +1,104 @@
...
+    address: Address,
+}
+
+impl WalletConnect {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        let client = Client::new(
+            "rad-ens",
+            Metadata {
+                description: "Change ENS records for your Radicle profile".into(),
+                url: "https://github.com/radicle-dev".parse()?,
```
cloudhead, line 21:
```suggestion
                url: "https://radicle.network".parse()?,
```

common/src/ethereum/walletconnect.rs
```diff
@@ -0,0 +1,104 @@
...
+                icons: vec!["https://avatars0.githubusercontent.com/u/48290027".parse()?],
```
cloudhead, line 22:
```suggestion
                icons: vec!["https://app.radicle.network/logo.png".parse()?],
```

common/src/ethereum/walletconnect.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct WalletConnect {
+    client: Client,
+    chain_id: u64,
+    address: Address,
+}
+
+impl WalletConnect {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
+        let client = Client::new(
+            "rad-ens",
```
cloudhead, line 18: What's this? :thinking: Profile? Any idea what this is supposed to be?

common/src/ethereum/walletconnect.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    fn address_string(&self) -> String {
+        format!("{}", self.address)
+    }
+
+    pub async fn sign_message<S: Send + Sync + AsRef<[u8]>>(
+        &self,
+        msg: S,
+    ) -> Result<Signature, CallError> {
+        let msg = unsafe { std::str::from_utf8_unchecked(msg.as_ref()) };
```
cloudhead, line 65: Can we take a `&str` or `String` as `msg` instead of having to convert it here?

common/src/ethereum/walletconnect.rs
```diff
@@ -0,0 +1,104 @@
...
+        )?;
+
+        Ok(WalletConnect {
+            client,
+            chain_id: 0,
+            address: Address::zero(),
+        })
+    }
+
+    pub async fn show_wait_connection(mut self) -> Result<Self, SessionError> {
```
cloudhead, line 34:
```suggestion
    pub async fn show_qr(mut self) -> Result<Self, SessionError> {
```
Clearer to me

common/src/ethereum.rs
xphoniex: Signing and sending both happens inside `ethers` which we can't intercept ([code](https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/signer.rs#L256-L264)).

common/src/ethereum/walletconnect.rs
xphoniex: This is the profile that's shown when walletconnect wants permission to connect with your app.

Screenshot from 2022-03-28 12-21-29

common/src/ethereum/walletconnect.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/0028efb90cc8297feef23f19dc3f55cd071afb6e#diff-363b2cf3885221c5b26bfa5862fc3e2363454ef177a777c5e11759468b8edc5cR34

common/src/ethereum/walletconnect.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/0028efb90cc8297feef23f19dc3f55cd071afb6e#diff-363b2cf3885221c5b26bfa5862fc3e2363454ef177a777c5e11759468b8edc5cR19-R24

common/src/ethereum/walletconnect.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/0028efb90cc8297feef23f19dc3f55cd071afb6e#diff-363b2cf3885221c5b26bfa5862fc3e2363454ef177a777c5e11759468b8edc5cR19-R24

common/src/ethereum/walletconnect.rs
xphoniex: Don't think so, tried at first but couldn't get it to compile.

common/src/ethereum/walletconnect.rs
cloudhead: But where does the string `rad-ens` show in this case?

common/src/ethereum/walletconnect.rs
cloudhead: But `msg: &str` here, so can't you just take `&str` as the function param?

common/src/ethereum/walletconnect.rs
cloudhead: If we don't know, maybe set it to `radicle-cli` :shrug:

common/src/ethereum.rs
```diff
@@ -223,6 +243,14 @@ impl Wallet {
                 .context("Could not connect to Ledger device")?;
 
             Ok(Wallet::Ledger(signer))
+        } else if options.walletconnect {
+            let signer = WalletConnect::new()
+                .map_err(|_| anyhow!(""))
+                .context("Failed to create WalletConnect client")?
```
cloudhead, line 249: Perhaps move this error string into the `anyhow!` right above, no need for `context()` then.

common/src/ethereum.rs
```diff
@@ -18,12 +18,17 @@ use ethers::types::Chain;
 use rad_terminal::args;
 use rad_terminal::components as term;
 
+mod walletconnect;
+
+use self::walletconnect::WalletConnect;
```
cloudhead, line 23:
```suggestion
use walletconnect::WalletConnect;
```

xphoniex:
```
error[E0659]: `walletconnect` is ambiguous
  --> common/src/ethereum.rs:23:5
   |
23 | use walletconnect::WalletConnect;
   |     ^^^^^^^^^^^^^ ambiguous name
   |
   = note: ambiguous because of multiple potential import sources
   = note: `walletconnect` could refer to a crate passed with `--extern`
   = help: use `::walletconnect` to refer to this crate unambiguously
note: `walletconnect` could also refer to the module defined here
  --> common/src/ethereum.rs:21:1
   |
21 | mod walletconnect;
   | ^^^^^^^^^^^^^^^^^^
   = help: use `self::walletconnect` to refer to this module unambiguously
   ```

File: common/src/ethereum/walletconnect.rs (hunk 9 above)
Comment (xphoniex) [reply common/src/ethereum/walletconnect.rs:65]: You mean like this?

```
    pub async fn sign_message(&self, msg: &str) -> Result<Signature, CallError> {
        self.client
            .personal_sign(&[msg, &self.address_string()])
            .await
    }
```

Issue is downstream:
```
error[E0308]: mismatched types
   --> common/src/ethereum.rs:204:54
    |
197 |     async fn sign_message<S: Send + Sync + AsRef<[u8]>>(
    |                           - this type parameter
...
204 |             Self::WalletConnect(s) => s.sign_message(message).await.map_err(WalletError::from),
    |                                                      ^^^^^^^ expected `&str`, found type parameter `S`
    |
    = note:   expected reference `&str`
            found type parameter `S`
note: return type inferred to be `&str` here
   --> common/src/ethereum.rs:200:41
    |
200 |       ) -> Result<Signature, Self::Error> {
    |  _________________________________________^
201 | |         match self {
202 | |             Self::Ledger(s) => s.sign_message(message).await.map_err(WalletError::from),
203 | |             Self::Local(s) => s.sign_message(message).await.map_err(WalletError::from),
204 | |             Self::WalletConnect(s) => s.sign_message(message).await.map_err(WalletError::from),
205 | |         }
206 | |     }
    | |_____^
```

common/src/ethereum.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ce9ad2d97fec16a5ef720cbb82ae87a96f6f5e28#diff-799f6adcb3a8a0469e7950f80331d4ffcb715140d420d1f4b25bad88ee61adc9R248

common/src/ethereum/walletconnect.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ce9ad2d97fec16a5ef720cbb82ae87a96f6f5e28#diff-363b2cf3885221c5b26bfa5862fc3e2363454ef177a777c5e11759468b8edc5cR18

common/src/ethereum.rs
cloudhead: Ah right, we are also importing a crate of the name, gotcha :)

common/src/ethereum/walletconnect.rs
cloudhead: Oh I see. Ok, I'll look into this later.

cloudhead (dismissed):
Nice work.

cloudhead:
Can you resolve the conflict (it's an easy one) and rebase, and then we can merge. I'll fix the dependency on walletconnect-rs after.

PR #61, opened 2022-03-25
rad-auth: Adds username & password to argument list

erak:
### Description

-  In order to make unit and integration testing possible, a username and password can now be passed: `rad-auth --username "Alice" --password "secret"`
- A first unit test was added, which depends on some new convenience functions. They should help to inspect local profiles and harness testing in general


common/src/profile.rs
```diff
@@ -21,11 +21,27 @@ pub fn default() -> Result<Profile, Error> {
...
+    
+    Ok(config.user_name()?)
+}
+
 /// List all profiles.
 pub fn list() -> Result<Vec<Profile>, Error> {
     lnk_profile::list(None).map_err(|e| e.into())
 }
 
+/// Get the count of all projects.
```
cloudhead, line 38:
```suggestion
/// Get the count of all profiles.
```

auth/src/lib.rs
```diff
@@ -188,3 +235,54 @@ pub fn authenticate(profiles: &[profile::Profile], options: Options) -> anyhow::
...
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn profile_can_be_initialized() -> Result<()> {
+        let home_dir = TempDir::new("lnk-home")?;
+        let home = home_dir.path().to_str().unwrap();
+        
+        temp_env::with_var("LNK_HOME", Some(home), || {
```
cloudhead, line 271: You can use `librad::profile::LNK_HOME` here and below to make the code a little more robust

auth/src/lib.rs
```diff
@@ -45,14 +51,47 @@ impl Args for Options {
...
+                        .value()?
+                        .to_str()
+                        .ok_or(anyhow::anyhow!(
+                            "invalid username specified with `--username`"
+                        ))?
+                        .to_owned();
+                    
+                    username = Some(val);
+                }
+                Long("password") if password.is_none() => {
```
cloudhead, line 65: You could add `init` as an additional condition, so that you have to pass `--init` first, otherwise password and username don't make sense.

cloudhead (dismissed):

PR #62, opened 2022-03-25
Add `rad gov`

xphoniex:
This won't be 100% compatible with the one from `radgov` due to use of different libraries.

Differences:

```
0xba000000000000000000000000000000000000be 0 "f2(bool)" 0x0000000000000000000000000000000000000000000000000000000000000001
0xca000000000000000000000000000000000000fe 0 "f3(uint256,bytes)" 1 0x1234
0xde000000000000000000000000000000000000ed 0 "f4(uint256[],bool[])" "[1, 2]" "[true, false]"
```

- value should be decimal, won't parse hex
- we can't pass hex to `f2` which is looking for `bool`, should be `false|true or 1|2`.
- in `f3`, bytes part can't start with `0x`
- in `f4`, there should be no space between args, e.g. `"[1, 2]" -> "[1,2]" `

also I'm using [`quoteless_string`](https://github.com/radicle-dev/radicle-cli/pull/62/files#diff-bbfe9cdef108fb3bcaf0ca04f1c727ec226ddebeb01454f2275aace4098a0d74R244-R248) to remove double quotes around tokens.


common/src/ethereum.rs
```diff
@@ -286,3 +286,38 @@ pub fn chain_from_id(id: u64) -> Option<Chain> {
 pub fn hex(bytes: impl AsRef<[u8]>) -> String {
     format!("0x{}", hex::encode(bytes))
 }
+
+/// Access the wallet specified in SignerOptions
+pub async fn get_wallet(
+    signer_opts: SignerOptions,
+    provider: Provider<Http>,
+) -> anyhow::Result<(Wallet, Provider<Http>)> {
+    use rad_terminal::args::Error;
```
cloudhead, line 295:
```suggestion
    use rad_terminal::args::Error;
    
```

gov/src/governance.rs
```diff
@@ -0,0 +1,80 @@
...
+    Provider(#[from] ProviderError),
+    #[error(transparent)]
+    Contract(#[from] ContractError<M>),
+    #[error(transparent)]
+    Abi(#[from] ethers::abi::Error),
+    #[error(transparent)]
+    SignerMiddleware(#[from] SignerMiddlewareError<Provider<Http>, ethereum::Wallet>),
+}
+
+type Proposal = (Address, U256, U256, U256, U256, U256, bool, bool);
```
cloudhead, line 37: This doesn't look right, ie. I think it'll fail when you `get_proposal()`: https://github.com/radicle-dev/radicle-contracts/blob/master/contracts/Governance/Governor.sol#L73

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+Options
+
+    --help  Print help
+
+Wallet options
+
+    --rpc-url <url>              JSON-RPC URL of Ethereum node (eg. http://localhost:8545)
+    --ledger-hdpath <hdpath>     Account derivation path when using a Ledger hardware device
+    --keystore <file>            Keystore file containing encrypted private key (default: none)
+    --walletconnect              WalletConnect account
```
cloudhead, line 43: Does this patch rely on the walletconnect one? If so, we need to merge that one first I guess, otherwise you can remove this.

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    Propose,
+    Vote,
+}
+
+pub struct Options {
+    provider: ethereum::ProviderOptions,
+    signer: ethereum::SignerOptions,
+    command: Command,
+    proposal_id: Option<U256>,
+    proposal_file: Option<OsString>,
```
cloudhead, line 61: Do you want to fold these into the enum maybe? Might be cleaner, eg.:

```rust
enum Command { Propose { proposal: OsString }, Vote { proposal_id: U256 } }
```

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        term::format::secondary(proposal.0),
+        format!("↑{}", term::format::highlight(proposal.4)),
+        format!("↓{}", term::format::warn(proposal.5)),
+        term::format::secondary(proposal.3),
+    ]);
+    term::blank();
+    table.render();
+    term::blank();
+
+    if let Some(vote) = term::select(&["true", "false"], &"true") {
```
cloudhead, line 235: Maybe "approvet" and "reject" is clearer?

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    Ok(())
+}
+
+async fn run_vote<M>(id: U256, governance: Governance<M>) -> anyhow::Result<()>
+where
+    M: Middleware + 'static,
+    crate::governance::Error<M>: From<<M as Middleware>::Error>,
+{
+    let proposal = governance.get_proposal(id).await?.unwrap();
```
cloudhead, line 217: We can do `.await??` here maybe to return a nicer error in case.

terminal/src/lib.rs
```diff
@@ -385,6 +385,7 @@ pub mod components {
     pub fn secret_input_with_prompt(prompt: &str) -> SecUtf8 {
         SecUtf8::from(
             Password::with_theme(&theme())
+                .allow_empty_password(true)
```
cloudhead, line 388: Let's not include this here

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    table.push([
+        term::format::bold("proposer"),
+        term::format::bold("for"),
+        term::format::bold("against"),
+        term::format::bold("end block"),
+    ]);
+    table.push([
+        term::format::secondary(proposal.0),
+        format!("↑{}", term::format::highlight(proposal.4)),
+        format!("↓{}", term::format::warn(proposal.5)),
```
cloudhead, line 228: For green/red in this case we can create a `format::positive` (green) and `format::negative` (red), I think it'll be clearer.

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
+use std::ffi::OsString;
+use std::fs::read_to_string;
```
cloudhead, line 2: I prefer not to import functions directly, as it takes extra effort to figure out on the callsite if it's a std function or a local function, so I usually do 

``````rust
use std::fs;
...
fs::read_to_string
```

gov/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        values.push(U256::from_dec_str(value).unwrap());
+        signatures.push(sig.to_string());
+        calldatas.push(calldata.clone());
+    }
+    spinner.finish();
+
+    let spinner = term::spinner("Sending proposal...");
+    let call = governance.propose(targets, values, signatures, calldatas, content)?;
+    ethereum::transaction(call).await?;
+    spinner.finish();
```
cloudhead, line 207: I think using a spinner here won't work well since `ethereum::transaction` does its own output..

cloudhead:
Nice work! I think the incompatibilities are fine. Maybe the bytes not starting with 0x is the only surprise, but we can look into that later.

gov/src/governance.rs
xphoniex: It's only returning primitive types, arrays and mappings are missing. You can run it against a proposal id and check locally.

gov/src/lib.rs
xphoniex: Actually we don't need `Option` here.

terminal/src/lib.rs
xphoniex: Hm.. I'm using a password-less local keystore and need this, any reason why we can't include this? It is in its own separate commit, for future reference.

gov/src/lib.rs
xphoniex: Done

terminal/src/lib.rs
cloudhead: Ok, I guess it allows users to have lower security if they wish, but we can eventually improve on this later.

PR #65, opened 2022-03-29
rad-init: Adds name, description and branch to options

erak:


geigerzaehler:
This looks good for our use-case in Upstream. Thanks!

init/src/lib.rs
```diff
@@ -22,12 +22,19 @@ Usage
...
+    --help               Print help
 "#,
 };
 
+#[derive(Default)]
 pub struct Options {
     pub path: Option<PathBuf>,
+    name: Option<String>,
+    description: Option<String>,
+    branch: Option<String>,
```
cloudhead, line 37: These can all be public.

init/src/lib.rs
```diff
@@ -37,8 +44,42 @@ impl Args for Options {
...
+                        .value()?
+                        .to_str()
+                        .ok_or(anyhow::anyhow!(
+                            "invalid project description specified with `--description`"
+                        ))?
+                        .to_owned();
+
+                    description = Some(value);
+                }
+                Long("branch") if branch.is_none() => {
```
cloudhead, line 74: Should be `"default_branch"` like in the usage instructions

init/src/lib.rs
```diff
@@ -37,8 +44,42 @@ impl Args for Options {
...
+                    let value = parser
+                        .value()?
+                        .to_str()
+                        .ok_or(anyhow::anyhow!("invalid branch specified with `--branch`"))?
```
cloudhead, line 78: Same here

cloudhead (dismissed):

jgresham:
awesome, thank you!

PR #69, opened 2022-04-04
rad-common: Adds initial data structure for proposals

erak:


common/src/proposal.rs
```diff
@@ -0,0 +1,190 @@
...
+    // a commit hash generated by git while creating a proposal.
+    pub id: Id,
+    // The title should contain a very short and concise summary of
+    // this proposal.
+    pub title: String,
+    // Description or "cover letter" of this proposal. The content is typed
+    // and can be one of the types defined by the `ContentType`.
+    pub description: Option<Text>,
+    // List of revisions (commits) included in this proposal. The field `head`
+    // of this `NonEmpty` object is the most recent revision.
```
cloudhead, line 42: Perhaps use docstrings here so they show up in rustdoc.

common/src/proposal.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Creates an empty proposal referencing a zeroed-out OID.
+impl Default for Metadata {
+    fn default() -> Self {
+        Metadata {
+            id: Id::from(git2::Oid::zero()),
+            title: "".to_owned(),
+            description: Some(Text {
+                content: "".to_owned(),
+                mime: ContentType::Plain,
+            }),
```
cloudhead, line 55: Perhaps the default description should be `None`

common/src/proposal.rs
```diff
@@ -0,0 +1,190 @@
...
+    // of this `NonEmpty` object is the most recent revision.
+    pub revisions: NonEmpty<Revision>,
+}
+
+/// Creates an empty proposal referencing a zeroed-out OID.
+impl Default for Metadata {
+    fn default() -> Self {
+        Metadata {
+            id: Id::from(git2::Oid::zero()),
+            title: "".to_owned(),
```
cloudhead, line 51: `String::default()`

common/src/proposal.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Commit hash
+type Revision = librad::git_ext::Oid;
+pub type Id = librad::git_ext::Oid;
+
+/// Content types supported by text fields in proposal data.
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum ContentType {
+    Plain,
+    Markdown,
```
cloudhead, line 14: I'd argue these two types are the same type :thinking:

PR #76, opened 2022-04-08
rad-patch: Create and list patches

erak:


common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+    S: AsRef<ReadOnly>,
+{
+    let storage = storage.as_ref();
+    let mut patches: Vec<Metadata> = vec![];
+    let master = repo
+        .resolve_reference_from_short_name(&format!("rad/{}", project.default_branch))?
+        .target()
+        .unwrap();
+
+    if let Ok(refs) = Refs::load(&storage, &project.urn, peer) {
```
erak, line 70: @cloudhead Is this the correct way to do this? It feels kind of complicated and I'm not 100% sure I covered all cases.

common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+    peer: Option<PeerId>,
+    project: &project::Metadata,
+    patch: &Metadata,
+    table: &mut term::Table<2>,
+) -> anyhow::Result<()>
+where
+    S: AsRef<ReadOnly>,
+{
+    let storage = storage.as_ref();
+    let peer_id = peer.unwrap_or(*storage.peer_id());
```
cloudhead, line 138: Could we instead use the `peer_id` in `Metadata`? Kind of strange to pass in a peer into this function, unless I'm missing something.

patch/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+    name: "patch",
+    description: env!("CARGO_PKG_DESCRIPTION"),
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad patch [<option>...]
+
+Options
+    --list    Prints all patches (default: false)
```
cloudhead, line 24:
```suggestion
    --list    List all patches (default: false)
```

patch/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+        term::blank();
+
+        let message = [title, description].join("\n");
+        patch::create(repo, &message, true)?;
+
+        if term::confirm_with_default("Sync to seed?", true) {
+            let sync_options = rad_sync::Options {
+                origin: None,
+                seed: None,
+                refs: rad_sync::Refs::DefaultBranch,
```
cloudhead, line 197: I think we want to sync the current branch here, not the default branch, eg. say you're proposing a commit on `feature/1`, that's the branch you'd want to sync. Have you tested on non-default branches?

terminal/src/lib.rs
```diff
@@ -202,6 +205,20 @@ pub mod components {
...
+            .language("md")
+            .grid(true)
+            .header(true)
+            .input(
+                BatInput::from_bytes(content.as_bytes())
+                    .name("tmp.md")
+                    .kind("File"),
+            )
+            .print()
+            .unwrap();
```
cloudhead, line 219: Since `bat` brings in lots of dependencies, I think it would be better to use the system `bat`, ie. check if the `bat` command exists, and if so, use it, otherwise print without highlighting.

cloudhead (commented):
Nice work. See feedback.

common/src/patch.rs
cloudhead: This does seem complicated, will have to try to understand what you're trying to do here and get back to you.

common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+use librad::git::refs::Refs;
+use librad::git::storage::{ReadOnly, ReadOnlyStorage};
+use librad::PeerId;
+
+use crate::{git, person, project};
+use rad_terminal::components as term;
+
+/// A patch can either be open or merged. It's considered merged, as soon as the
+/// merge-base of the default's branch HEAD and the patch commit is the same than the
+/// patch commit itself.
```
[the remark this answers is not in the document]
cloudhead: After reading more of the code, I think I understand this, but am worried it's not exactly what we want, ie. as soon as another commit is added on the merge base, the patch will be back to being open?

common/src/patch.rs
cloudhead: So basically here we're loading the `signed_refs`, so any ref that was pushed to the monorepo and signed, then we're getting the tag oids, then for each oid we're getting the tag, and getting the merge base between the tag's commit and the repo's default branch. I think this makes sense?

common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Head commit that the author wants to merge with this patch.
+    pub commit: Oid,
+    /// The merge base of [`Metadata::commit`] and the head commit of the first maintainer's default
+    /// branch.
+    pub merge_base: Option<Oid>,
+}
+
+impl Metadata {
+    pub fn state(&self) -> State {
+        if self.merge_base.unwrap() == self.commit {
```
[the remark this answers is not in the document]
cloudhead: With git you can do `git branch --contains <oid>` to list all branches that contain <oid>. With git2 I'm not sure what the best way is, but you can use `parents` to get all parent commits of head, and check if one of those commits is the patch, though that'll be expensive for long histories..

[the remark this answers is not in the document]
cloudhead: But note that if it was merged via a merge commit, this won't work!

common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(PartialEq, Eq)]
+pub enum State {
+    Open,
+    Merged,
+}
+
+/// A patch is a change set that a user wants the maintainer to merge into a projects default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
```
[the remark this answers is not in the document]
rudolfs: We're using `radicle-patch/` in Upstream.

terminal/src/lib.rs
erak: Ok, switched to shelling out to `bat` (if available.)

common/src/patch.rs
erak: Changed the function and it's signature quite a bit.

common/src/patch.rs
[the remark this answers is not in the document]
erak: Removed this entirely such that the merge base is not stored anymore with the patch. Checking it now on demand, whenever needed.

common/src/patch.rs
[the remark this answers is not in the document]
erak: @cloudhead @rudolfs @sebastinez When would be a good time for the sync CLI + Ustream + Web? Would stay with `radicle-patch/` till then.

common/src/patch.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// ID of a patch. This is the portion of the tag name following the `radicle-patch/` prefix.
+    pub id: String,
+    /// Peer that the patch originated from
+    pub peer: PeerId,
+    /// Message attached to the patch. This is the message of the annotated tag.
+    pub message: Option<String>,
+    /// Head commit that the author wants to merge with this patch.
+    pub commit: Oid,
+    /// The merge base of [`Metadata::commit`] and the head commit of the first maintainer's default
+    /// branch.
```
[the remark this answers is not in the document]
erak: Removed that altogether.

common/src/patch.rs
erak: Yes, I want to load the `signed_refs` but needed to be able to get a typed tag in the end. And the only way I could figure out was by going through the target object and then peeling a tag out :smirk: 
Is there a better way to do this?

common/src/patch.rs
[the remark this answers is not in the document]
erak: Simplified this again.

common/src/patch.rs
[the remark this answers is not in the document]
rudolfs: Any time would work for me really. Can also discuss it in #alt-clients discord if it's only about the paths.

common/src/patch.rs
```diff
@@ -0,0 +1,191 @@
...
+        for blob in blobs {
+            match storage.find_object(blob.1) {
+                Ok(Some(object)) => {
+                    let tag = object.peel_to_tag()?;
+                    patches.push(Metadata {
+                        id: tag.name().unwrap().to_string(),
+                        peer: info.clone(),
+                        message: Some(tag.message().unwrap().to_string()),
+                        commit: tag.target_id(),
+                    });
```
cloudhead, line 75: Shouldn't we check whether the tag is a patch, eg. by checking the tag name?

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    if let Ok(refs) = Refs::load(&storage, &project.urn, peer_id) {
+        let blobs = match refs {
+            Some(refs) => refs.tags().collect(),
+            None => vec![],
+        };
+        for (_, oid) in blobs {
+            match storage.find_object(oid) {
+                Ok(Some(object)) => {
+                    let tag = object.peel_to_tag()?;
```
erak, line 68: Calling `object.into_tag` results in the compiler error `libgit2_sys::git_repository cannot be shared between threads safely`. Shall we look into this seperately, since I could not resolve that easily?

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+pub fn self_info<S>(storage: &S, project: &project::Metadata) -> Result<project::PeerInfo>
+where
+    S: AsRef<ReadOnly>,
+{
+    let storage = storage.as_ref();
+    let urn = storage.config()?.user()?.unwrap();
+    let peer_id = storage.peer_id();
+    let name = person::get(storage, &urn)?.map(|p| p.subject().name.to_string());
```
erak, line 191: Calling `person::local` needs a non-`ReadOnly` storage. And since passing a `ReadOnly` storage into `self_info` makes sense semantically, I'd lean towards leaving the code as is. Otherwise we'd need to get the writeable storage using a signer etc, which would result in even more code, I guess.

common/src/patch.rs
cloudhead: Yup.

common/src/patch.rs
cloudhead: That's a bug in `person::local` then I guess, but let's solve it another day.

terminal/src/lib.rs
```diff
@@ -273,6 +276,18 @@ pub mod components {
...
     }
 
+    pub fn markdown(content: &str) {
+        if !content.is_empty() {
+            blank();
+            if command::bat(["-p", "-l", "md"], content).is_err() {
+                blob(content);
+
+                blank();
+                warning("Could not apply syntax highlighting. Please make sure `bat` is available.")
```
cloudhead, line 286: I think we decided to remove the warning?

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub peer: project::PeerInfo,
+    /// Message attached to the patch. This is the message of the annotated tag.
+    pub message: Option<String>,
+    /// Head commit that the author wants to merge with this patch.
+    pub commit: Oid,
+}
+
+impl Metadata {
+    pub fn name(&self) -> String {
+        str::replace(&self.id, TAG_PREFIX, "")
```
cloudhead, line 37: `strip_prefix` would be better, since the tag prefix could occur in other parts of the string, and we wouldn't want to remove it there.

cloudhead: Actually the id shouldn't have the prefix, so I don't think this function should exist..

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+    Open,
+    Merged,
+}
+
+/// A patch is a change set that a user wants the maintainer to merge into a project's default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
+pub struct Metadata {
+    /// ID of a patch. This is the portion of the tag name following the `radicle-patch/` prefix.
```
cloudhead, line 25:
```suggestion
    /// ID of a patch. This is the portion of the tag name following the `patches/` prefix.
```

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+#[derive(PartialEq, Eq)]
+pub enum State {
+    Open,
+    Merged,
+}
+
+/// A patch is a change set that a user wants the maintainer to merge into a project's default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
```
cloudhead, line 23:
```suggestion
/// A patch is represented by an annotated tag, prefixed with `patches/`.
```

common/src/patch.rs
```diff
@@ -0,0 +1,219 @@
...
+    Ok(())
+}
+
+/// Tries to construct a patch from ['git2::Tag'] and ['project::PeerInfo'].
+/// If the tag name matches the radicle patch prefix, a new patch metadata is
+/// created.
+pub fn from_tag(tag: git2::Tag, info: project::PeerInfo) -> Result<Option<Metadata>> {
+    let patch = match tag.name() {
+        Some(id) => {
+            let prefix = Regex::new(&format!("{}*", TAG_PREFIX)).unwrap();
```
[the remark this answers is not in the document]
erak: Thanks :roll_eyes:

terminal/src/lib.rs
erak: Dang, that should not be there anymore. Thanks

cloudhead (approved):
:tada:

cloudhead:
I made some major refactors to the cli libs and for some reason thought this PR was merged already, so now I'm fixing conflicts and will force push here. Unfortunately I've had to squash some of the history, but no biggie.

PR #78, opened 2022-04-12
`rad-inspect`: Add option to view `--history`

xphoniex:


xphoniex:
> Can we propagate errors instead of unwrapping?

Which ones?

xphoniex:
> > > Can we propagate errors instead of unwrapping?
> > 
> > 
> > Which ones?
> 
> All of them :)

Places where `.unwrap()` is still around are places which are not really supposed to fail e.g.:

```
                    if tip.parent_count() == 0 {
                        break;
                    }

                    tip = tip.parent(0).unwrap(); // we know parent exists because of count > 0
```

Or here, we could be in a case where commit has no `tree`, when it's never been updated:
```
                    let tree = tip.tree()?; // if `tree` does not exist, we exit here, rest are smooth sailing...
                    let entry = tree.get(0).unwrap().id();
                    let blob = storage
                        .find_object(Box::new(entry))?
                        .unwrap()
                        .into_blob()
                        .unwrap();
```

But once we're past that, we know for sure that that `tree` has 2 entries, and blobs can be read. Only time these would fail is when monorepo has been corrupted. Do you still think I should account for those cases?

inspect/src/lib.rs
```diff
@@ -127,6 +138,60 @@ pub fn run(options: Options) -> anyhow::Result<()> {
...
+                let mut tip = reference.peel_to_commit()?;
+
+                loop {
+                    let tree = tip.tree()?;
+                    let entry = tree.get(0).unwrap().id();
+                    let blob = storage
+                        .find_object(Box::new(entry))?
+                        .unwrap()
+                        .into_blob()
+                        .unwrap();
```
cloudhead, line 154: This one would fail if the object is not a blob, which is unlikely, but possible.

inspect/src/lib.rs
```diff
@@ -127,6 +138,60 @@ pub fn run(options: Options) -> anyhow::Result<()> {
...
+        match storage.reference(&branch) {
+            Ok(Some(reference)) => {
+                let mut tip = reference.peel_to_commit()?;
+
+                loop {
+                    let tree = tip.tree()?;
+                    let entry = tree.get(0).unwrap().id();
+                    let blob = storage
+                        .find_object(Box::new(entry))?
+                        .unwrap()
```
cloudhead, line 152: This one can fail for any i/o error, so definitely I'd handle it.

inspect/src/lib.rs
```diff
@@ -127,6 +138,60 @@ pub fn run(options: Options) -> anyhow::Result<()> {
...
         );
+    } else if options.history {
+        let branch = Reference::try_from(&urn)?;
+        match storage.reference(&branch) {
+            Ok(Some(reference)) => {
+                let mut tip = reference.peel_to_commit()?;
+
+                loop {
+                    let tree = tip.tree()?;
+                    let entry = tree.get(0).unwrap().id();
```
cloudhead, line 149: This one fails if the tree is empty, unlikely but possible.

terminal/src/lib.rs
```diff
@@ -73,6 +73,37 @@ pub mod components {
...
+    pub struct Box(pub String);
+
+    impl fmt::Display for Box {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            let mut width = self
+                .0
+                .lines()
+                .map(console::measure_text_width)
+                .max()
+                .unwrap()
```
cloudhead, line 85: This can fail with an empty string no?

inspect/src/lib.rs
```diff
@@ -127,6 +138,60 @@ pub fn run(options: Options) -> anyhow::Result<()> {
...
+                            term::format::highlight(term::format::bold(time)),
+                            serde_json::to_string_pretty(&content)?.to_colored_json_auto()?,
+                        ))
+                    );
+
+                    if tip.parent_count() == 0 {
+                        break;
+                    }
+
+                    tip = tip.parent(0).unwrap();
```
cloudhead, line 184: Why not combine this parent lookup with the `break` above

xphoniex: wdym?

xphoniex: https://github.com/radicle-dev/radicle-cli/pull/78/commits/205ece9cbc8d2ea54cf213f892f2a8f1e71ce4ff#diff-68a7e8c16c74cba4f71df82b4d89eac330072db4fcb43a4bd7de8bf3163283d9R185-R188

cloudhead (dismissed):

cloudhead:
Nice.

Issue #77, opened 2022-04-11
`rad auth` should create key and identity if profile is empty


geigerzaehler:
`rad auth` should be able to handle the case where a profile including a monorepo exists but no key and local identity has been created yet. This situation occurs with Upstream at the moment. This would also guard against the case where the user aborts `rad auth` when asked to type in their passphrase.

erak:
Hey @geigerzaehler, I will have a look and hoping to also fix #72 in one go.

sebastinez:
Is it possible that this issue is related with #81?

erak:
@sebastinez Yes, it is somewhat related I guess, but only in terms of handling errornous profiles. Issue #81 should not happen, whereas the empty profile created by `upstream-proxy` will be there in case the user starts up Upstream before using `rad auth` for the first time (@geigerzaehler hope to got this right though)

Since I wanted to unblock Upstream first, I decided to separate both issues for now.

geigerzaehler:
> whereas the empty profile created by `upstream-proxy` will be there in case the user starts up Upstream before using `rad auth` for the first time (@geigerzaehler hope to got this right though)

You’re spot on

Issue #72, opened 2022-04-07
Warn if LNK_HOME contains invalid profile


erak:
When calling `rad auth` in that case, profiles can't be loaded and CLI suggests to just run `rad auth`:
```
$ rad auth
== Authentication failed failed to load radicle profile
To setup your radicle profile, run `rad auth`.
```

PR #82, opened 2022-04-23
rad-auth: handle empty and non-active profiles

erak:
Fixes https://github.com/radicle-dev/radicle-cli/issues/77.
Fixes https://github.com/radicle-dev/radicle-cli/issues/72.

This PR changes the `rad auth` behaviour such that it handles empty (non readable monorepo) and non-active profiles:
- if empty profile(s) exists (e.g. the one created by `upstream-proxy`), these
  - a) issue a warning
  - b) are ignored when creating a new profile and identity
  - c) won't show up in profile selection anymore
- if a non-empty profile exists, but no active profile was found (deleted `active_profile` file), user can activate a fallback profile (first in the list)

### DIFF
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -189,7 +189,7 @@ pub fn authenticate(profiles: &[profile::Profile], options: Options) -> anyhow::
     }
 
     let selection = if profiles.len() > 1 && !options.active {
-        if let Some(p) = term::format::profile_select(profiles, &profile) {
+        if let Some(p) = term::format::profile::select(profiles, &profile) {
             p
         } else {
             return Ok(());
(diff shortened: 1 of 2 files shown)


cloudhead:
This seems like lots of additional complexity and i/o for handling a case that isn't supposed to happen. I would prefer if we just handled errors more gracefully than eagerly checking all profiles for issues.

erak:
@cloudhead Ok, got it. I created a more simple version: https://github.com/radicle-dev/radicle-cli/pull/83 which supersedes this PR.

erak:
Closing this in favor of https://github.com/radicle-dev/radicle-cli/pull/83.

PR #86, opened 2022-04-27
Speed up main CI job

adaszko:
[Before](https://github.com/radicle-dev/radicle-cli/actions/runs/2225929322): 30m 11s
[After](https://github.com/radicle-dev/radicle-cli/actions/runs/2232944858): 9m 30s

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -26,7 +26,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: cargo-${{ matrix.os }}-${{ hashFiles('**/Cargo.lock') }}
+          key: cargo-${{ matrix.os }}-fast-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -89,3 +71,18 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
+
+  docs:
```
adaszko, line 75: @cloudhead Do we even need `cargo doc` in CI?  I don't think it fails the CI job even if there are egregious syntax errors:

```
$ env RUSTFLAGS="-D warnings" cargo doc --profile ci --no-deps
[...]
warning: unresolved link to `fdsf`
 --> push/src/lib.rs:9:6
  |
9 | /// [`fdsf]
  |      ^^^^^ no item named `fdsf` in scope
  |
  = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
  = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

warning: `rad-push` (lib doc) generated 1 warning
    Finished ci [unoptimized] target(s) in 2m 22s
% echo $?
0
```

adaszko: Correction: I misread the code: It says `RUSTDOCFLAGS`, not `RUSTFLAGS` and it does return a non-zero exit code as is desired in CI.

cloudhead (approved):
:partying_face:

PR #89, opened 2022-04-29
Allow for piping password to rad-auth

adaszko:
Allows doing this, for easier password input:

```
pass radicle/auth | rad-auth
```

where `pass` is https://www.passwordstore.org/ or a similar command line-accessible password manager.

### DIFF
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -1,6 +1,7 @@
 use std::ffi::OsString;
 
 use anyhow::Context as _;
+use zeroize::Zeroizing;
 
 use librad::crypto::keystore::pinentry::SecUtf8;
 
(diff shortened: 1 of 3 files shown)


auth/src/lib.rs
```diff
@@ -7,6 +7,7 @@ use librad::crypto::keystore::pinentry::SecUtf8;
 use rad_common::{git, keys, person, profile};
 use rad_terminal::args::{Args, Error, Help};
 use rad_terminal::components as term;
+use zeroize::Zeroizing;
```
cloudhead, line 10: External deps should be grouped together, so this can go next to `anyhow`

cloudhead (commented):
Looks good, just one nit.

adaszko:
Just as a curiosity: On macOS, the built-in KeyChain can be used like so: `security find-generic-password -w -a $LOGNAME -s radicle-auth -w | rad-auth`

PR #91, opened 2022-05-02
Fix a few Ethereum & WalletConnect related issues

xphoniex:


ens/src/lib.rs
```diff
@@ -287,7 +287,7 @@ async fn setup(
         }
     }
 
-    let call = resolver.multicall(calls)?.gas(21000);
+    let call = resolver.multicall(calls)?;
```
cloudhead, line 291: You tested this? It wasn't working without previously no?

ens/src/lib.rs
```diff
@@ -290,27 +291,29 @@ async fn setup(
...
-        storage,
-    ) {
-        Ok(doc) => {
-            spinner.finish();
-            term::blob(serde_json::to_string(&doc.payload())?);
-        }
-        Err(err) => {
-            spinner.failed();
-            return Err(err);
+    if chain_id == 1 {
```
cloudhead, line 294: Makes sense but this function doesn't do anything else, so we should return an error instead.

ens/src/lib.rs
xphoniex: Yes, it's tested, this should work fine now.

ens/src/lib.rs
cloudhead: Also check the `Chain` type un the ethereum module to convert the uint.

common/src/ethereum.rs
```diff
@@ -80,6 +80,7 @@ impl SignerOptions {
                 }
                 Long("walletconnect") => {
                     options.walletconnect = true;
+                    std::env::set_var("RAD_SIGNER_LEGACY", "true");
```
cloudhead, line 83: Not sure this is the best way to pass a variable down :joy:

ens/src/lib.rs
xphoniex: Hm... it updates local identity. I don't want to pollute it when working with test chains.
Screenshot from 2022-05-02 17-32-01

common/src/ethereum/walletconnect.rs
```diff
@@ -12,6 +12,14 @@ pub struct WalletConnect {
     address: Address,
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum WalletError {
+    #[error(transparent)]
+    CallError(#[from] CallError),
+    #[error("failed to sign the tx")]
+    FailedToSignTx,
```
cloudhead, line 20: We usually don't include "error/failed" in the name of error variants, as it's redundant, so I'd suggest `WalletError::Call` and `WalletError::TransactionSignature`.

We should also improve the error message of the latter, as it'll be displayed to users, but I'm not sure in what case that would fail?

common/src/ethereum.rs
xphoniex: Correct, this is temporary, and we'll eventually omit this once we move to v2.0. Also would have to add a lot of code, every place which calls `ethereum::transaction` has to find out if wallet is `WalletConnect`. there are 6 places currently which means at least 18 more lines, and an unnecessary flag on `ethereum::transaction` signature. 

Made a compromise here.

ens/src/lib.rs
cloudhead: Yes I agree, but maybe include a warning then `Warning: Skipping local ENS setup`.

common/src/ethereum/walletconnect.rs
xphoniex: Fixed.

I noticed sometimes the raw response from API was lacking the signature part. Perhaps it was due to how spinner interferes with stdout, because once I disabled spinner, I wasn't able to reproduce this.

But regardless, if we ever fail to extract v,r,s from sig, we'll show that error which is more informative.

ens/src/lib.rs
xphoniex: Done

common/src/ethereum/signer.rs
```diff
@@ -0,0 +1,258 @@
...
+        &self,
+        data: T,
+        _: &Address,
+    ) -> Result<Signature, Self::Error> {
+        self.signer
+            .sign_message(data.into())
+            .await
+            .map_err(SignerMiddlewareError::SignerError)
+    }
+}
```
cloudhead, line 258: Why did we copy all of this over from ethers?

cloudhead: I mean why do we have our own Signer type with all the same methods plus one extra and our own SignerMiddleware and SignerMiddlewareError?

xphoniex: Ok, refactored it. Check [this](https://github.com/radicle-dev/radicle-cli/compare/171dad3c0b2bba1a7a257d712084e7ff83f23099..49ad2de82da8cd7a938e3448676583d6cc5d806a) please.

PR #97, opened 2022-05-09
Make patches work via HTTP

cloudhead:


patch/src/lib.rs
```diff
@@ -218,24 +215,20 @@ fn list_by_state(
...
 ) -> anyhow::Result<()> {
-    let mut patches: Vec<patch::Metadata> = list_patches(&storage, None, project)?;
+    let mut patches: Vec<patch::Metadata> = patch::all(project, None, &storage)?;
 
     for (_, info) in project::tracked(project, storage)? {
-        let mut theirs = list_patches(&storage, Some(info), project)?;
+        let mut theirs = patch::all(project, Some(info), &storage)?;
         patches.append(&mut theirs);
     }
+    patches.retain(|patch| state == patch::state(repo, patch));
```
erak, line 224: Ah, that's a lot better.

PR #98, opened 2022-05-10
Fix missing shared libraries problem on M1 Macs

adaszko:
Fix missing shared libraries problem on M1 Macs: libssl and libusb.

~~A workaround in libusb-sys has been implemented.  The bug is yet to be reported upstream to libusb.  It's to do with missing libobj.a static library on new versions of macOS.  libusb pkg-config .pc file seems to be assuming they're still there.~~

~~Once the issue is resolved upstream, the workaround in libusb-sys will be removed.~~

According to my understanding, the static build of libusb1-sys on macOS fails because of [a bug in rustc](https://github.com/rust-lang/rust/issues/96943).

IMO, it's best if we maintain a version of libusb1-sys with a workaround (referenced in this PR via `patch.io-crates`) until a rustc fix gets released.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,7 +104,7 @@ jobs:
         with:
           profile: minimal
       - name: Build the release binaries
-        run: cargo build --release
+        run: env OPENSSL_STATIC=1 OPENSSL_DIR=/usr/local/opt/openssl@1.1 LIBUSB_STATIC=1 cargo build --release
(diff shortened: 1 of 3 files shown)


adaszko:
I've also opened [an issue in libusb-sys](https://github.com/a1ien/rusb/pull/133).

cloudhead:
Woops, this doesn't build.

PR #99, opened 2022-05-11
Add rad-issue to commonly used commands

sebastinez:
This PR adds `rad-issue` to the listing of common used commands, when running `rad`


cloudhead:
I don't want people creating issues until the data layout is stable, otherwise we'll have issues..

cloudhead:
Let's re-open this when we're ready.

PR #102, opened 2022-05-17
common: Add ENS name to PeerIdentity and PeerInfo

sebastinez:
This PR tries to resolve from the `person.payload()` the `Ens` extension and adds it to the `PeerIdentity` and `PeerInfo`
We need this PR to merge for the issue listing, where we want to resolve a ENS name from a Radicle URN.

### DIFF
--- a/common/src/project.rs
+++ b/common/src/project.rs
@@ -32,6 +32,7 @@ use librad::PeerId;
 
 use lnk_identities;
 
+use crate::person::Ens;
 use crate::{git, person, seed};
 
 /// URL scheme for radicle resources.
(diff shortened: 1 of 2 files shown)


common/src/project.rs
```diff
@@ -115,18 +116,26 @@ impl TryFrom<Url> for Origin {
 pub struct PeerIdentity {
     pub urn: Urn,
     pub name: String,
+    pub ens: Option<String>,
```
cloudhead, line 119: We should keep this as `Ens`, there may be more things we want to include in the ens payload in the future than the name.

sebastinez: Okay no problem 👍

cloudhead (dismissed):

Issue #88, opened 2022-04-29
Verify authenticity of peer refs when fetching from seeds


geigerzaehler:
At the moment, `rad sync` does not verify the authenticity of references of remote peers it fetches from a seed. As a result seeds can manipulate references of a project which can be used to inject malicious code.

Specifically, `rad sync` should verify that all references for a project by a certain peer match the content of `rad/signed_refs` for that peer and the content itself is signed with the peer’s key.

PR #105, opened 2022-05-17
Verify signed refs on fetch

cloudhead:
Related to #88


geigerzaehler:
Wouldn’t it be safer to also delete invalid refs from the monorepo? Otherwise `git fetch rad` would still pick them up, right?

cloudhead:
> Wouldn’t it be safer to also delete invalid refs from the monorepo? Otherwise `git fetch rad` would still pick them up, right?

Yeah, you mean if refs were found to be invalid, delete them before exiting?

cloudhead:
Going to merge this, but won't close #88 until it's fully solved.

Issue #104, opened 2022-05-17
Add ability to edit project name and description


cloudhead:
Currently, there's no way to change a project after it is created. We should add the ability to edit the name and description, via an interface like this:

```shell
$ cd my-project/
$ rad edit --help
   rad-edit 0.1.0
   Edit project payloads

Usage

   rad edit [--project <urn>]

$ rad edit
  <editor opens>
  {"name": "my-project", "description": "blah blah"}
  <save>
[ok] Project successfully updated to <commit-hash>
```

The commmand open open your editor via the `term::Editor` type and allow you to edit the project payload and then save it.

cloudhead:
The plan is to have other commands for editing the other sections of the identity:
* Editing extensions should be done by extension-specific sub-commands
* Editing delegates should be done via a CRUD-like cli, ie. `rad delegate <add|remove>`

cloudhead:
This is done!

PR #110, opened 2022-05-18
rad-edit: Initial commit

xphoniex:
Should fix issue https://github.com/radicle-dev/radicle-cli/issues/104

### DIFF
--- a/edit/src/lib.rs
+++ b/edit/src/lib.rs
@@ -1,0 +1,117 @@
+use std::ffi::OsString;
+use std::str::FromStr;
+
+use radicle_common::args::{Args, Error, Help};
+use radicle_common::{keys, profile};
+use radicle_terminal as term;
+
+use librad::git::identities::{any, person, project, SomeIdentity};
+use librad::git::Urn;
+
+use link_identities::payload::{PersonPayload, ProjectPayload};
+
+use anyhow::anyhow;
+
+pub const HELP: Help = Help {
+    name: "edit",
+    description: env!("CARGO_PKG_DESCRIPTION"),
+    version: env!("CARGO_PKG_VERSION"),
+    usage: r#"
+Usage
+
+    rad edit [<urn>] [<option>...]
+
+    Edits the identity document pointed to by the URN. If it isn't specified,
+    the current project is edited.
+
+Options
+
+    --help              Print help
+"#,
+};
+
+#[derive(Default, Debug, Eq, PartialEq)]
+pub struct Options {
+    pub urn: Option<Urn>,
+}
+
+impl Args for Options {
+    fn from_args(args: Vec<OsString>) -> anyhow::Result<(Self, Vec<OsString>)> {
+        use lexopt::prelude::*;
+
+        let mut parser = lexopt::Parser::from_args(args);
+        let mut urn: Option<Urn> = None;
+
+        while let Some(arg) = parser.next()? {
+            match arg {
+                Long("help") => {
+                    return Err(Error::Help.into());
+                }
+                Value(val) if urn.is_none() => {
+                    let val = val.to_string_lossy();
+
+                    if let Ok(val) = Urn::from_str(&val) {
+                        urn = Some(val);
+                    } else {
+                        return Err(anyhow!("invalid URN '{}'", val));
+                    }
+                }
+                _ => return Err(anyhow::anyhow!(arg.unexpected())),
+            }
+        }
+
+        Ok((Options { urn }, vec![]))
+    }
+}
+
+pub fn run(options: Options) -> anyhow::Result<()> {
+    let profile = profile::default()?;
+    let signer = term::signer(&profile)?;
+    let storage = keys::storage(&profile, signer)?;
+
+    let urn = options
+        .urn
+        .or_else(|| radicle_common::project::cwd().ok().map(|(urn, _)| urn))
+        .ok_or_else(|| anyhow!("Couldn't get URN from either command line or cwd"))?;
+
+    let identity = any::get(&storage, &urn)?
+        .ok_or_else(|| anyhow!("No project or person found for this URN"))?;
+
+    match identity {
+        SomeIdentity::Project(_) => {
+            let payload = serde_json::to_string_pretty(
+                project::verify(&storage, &urn)?
+                    .ok_or_else(|| anyhow!("Couldn't get project's identity doc"))?
+                    .payload(),
+            )?;
+            match term::Editor::new().edit(&payload)? {
+                Some(updated_payload) => {
+                    let payload: ProjectPayload = serde_json::from_str(&updated_payload)?;
+                    project::update(&storage, &urn, None, payload, None)?;
+                }
+                None => return Err(anyhow!("Operation aborted!")),
+            }
+        }
+        SomeIdentity::Person(_) => {
+            let payload = serde_json::to_string_pretty(
+                person::verify(&storage, &urn)?
+                    .ok_or_else(|| anyhow!("Couldn't get person's identity doc"))?
+                    .payload(),
+            )?;
+            match term::Editor::new().edit(&payload)? {
+                Some(updated_payload) => {
+                    let payload: PersonPayload = serde_json::from_str(&updated_payload)?;
+                    person::update(&storage, &urn, None, payload, None)?;
+                }
+                None => return Err(anyhow!("Operation aborted!")),
+            }
+        }
+        _ => {
+            anyhow::bail!("Operation not supported for identity type of {}", urn)
+        }
+    };
+
+    term::success!("Update successful!");
+
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4678,6 +4678,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "rad-edit"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "lexopt",
+ "librad",
+ "link-identities",
+ "radicle-common",
+ "radicle-terminal",
+ "serde_json",
+]
+
+[[package]]
 name = "rad-ens"
 version = "0.1.0"
 dependencies = [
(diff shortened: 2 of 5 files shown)


edit/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    };
+
+    let identity = any::get(&storage, &urn)?
+        .ok_or_else(|| anyhow!("No project or person found for this URN"))?;
+
+    if let Some(updated_payload) = term::Editor::new()
+        .edit(&serde_json::to_string_pretty(&identity.payload())?)
+        .unwrap()
+    {
+        if identity.project().is_some() {
```
alexjg, line 88: This would probably be nicer as a match on `SomeIdentity` (which is what is returned by `any::get`).

FintanH:
I should probably clarify, I think it would be beneficial to first think about how all of the document should be updated, not just the payload, as mentioned in https://github.com/radicle-dev/radicle-cli/issues/104#issuecomment-1131436073.

I also think it would be worth using [`lnk-identities::project::update`](https://github.com/radicle-dev/radicle-link/blob/master/cli/lnk-identities/src/project.rs#L179-L219) for a couple of reasons:
* It updates the **verified** document, ie. the tip that all delegates agree on
* It will handle the addition of extensions correctly
* Has logic for adding delegations

edit/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    let profile = profile::default()?;
+    let signer = term::signer(&profile)?;
+    let storage = keys::storage(&profile, signer)?;
+
+    let urn = if let Some(urn) = options.urn {
+        urn
+    } else {
+        let repo = git::Repository::open(Path::new("."))?;
+
+        radicle_common::project::rad_remote(&repo)?.url.urn
```
cloudhead, line 78: `project::cwd()` does this for you

edit/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+        .unwrap()
+    {
+        if identity.project().is_some() {
+            let payload: ProjectPayload = serde_json::from_str(&updated_payload)?;
+            project::update(&storage, &urn, None, payload, None)?;
+        } else {
+            let payload: PersonPayload = serde_json::from_str(&updated_payload)?;
+            person::update(&storage, &urn, None, payload, None)?;
+        }
+        term::info!("Updated!");
```
cloudhead, line 95: `term::success("Update successful")`

edit/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+        term::warning("Aborted!");
```
erak, line 97: I think for most of the other commands, we're returning an error if the user cancels the current flow.

edit/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ec0b9679b7a5e43fc48691fca49f262e317765bb#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R91-R101

edit/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ec0b9679b7a5e43fc48691fca49f262e317765bb#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R74

edit/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ec0b9679b7a5e43fc48691fca49f262e317765bb#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R102

xphoniex: also, `success!(...)` is a macro.

Is there a logic behind which ones are fns e.g. `term::warning(...)` and the ones which are macro?

We should create macros for fns, just to make it simpler for devs to use.

edit/src/lib.rs
xphoniex: https://github.com/radicle-dev/radicle-cli/commit/ec0b9679b7a5e43fc48691fca49f262e317765bb#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R106

xphoniex:
> I should probably clarify, I think it would be beneficial to first think about how all of the document should be updated, not just the payload, as mentioned in [#104 (comment)](https://github.com/radicle-dev/radicle-cli/issues/104#issuecomment-1131436073).
> 
> I also think it would be worth using [`lnk-identities::project::update`](https://github.com/radicle-dev/radicle-link/blob/master/cli/lnk-identities/src/project.rs#L179-L219) for a couple of reasons:
> 
> * It updates the **verified** document, ie. the tip that all delegates agree on
> * It will handle the addition of extensions correctly
> * Has logic for adding delegations

Thanks Fintan, since we're planning on dealing with extensions and delegations elsewhere, the only thing remaining is the verified tip which I'm now retrieving [using](https://github.com/radicle-dev/radicle-cli/commit/ec0b9679b7a5e43fc48691fca49f262e317765bb#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R77-R88):

```
    let identity = any::get(&storage, &urn)?
        .ok_or_else(|| anyhow!("No project or person found for this URN"))?;

    let payload = match identity {
        SomeIdentity::Project(_) => {
            serde_json::to_string_pretty(project::verify(&storage, &urn)?.unwrap().payload())?
        }
        SomeIdentity::Person(_) => {
            serde_json::to_string_pretty(person::verify(&storage, &urn)?.unwrap().payload())?
        }
        _ => todo!(),
    };
```

This way we can use the simple `update` fns directly (without parsing two different structs from json) and we have access to verified heads. Comments?

edit/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+        .ok_or_else(|| anyhow!("No project or person found for this URN"))?;
+
+    let payload = match identity {
+        SomeIdentity::Project(_) => {
+            serde_json::to_string_pretty(project::verify(&storage, &urn)?.unwrap().payload())?
+        }
+        SomeIdentity::Person(_) => {
+            serde_json::to_string_pretty(person::verify(&storage, &urn)?.unwrap().payload())?
+        }
+        _ => todo!(),
```
FintanH, line 87: You may want to throw an error message here. Hitting this case would mean that radicle-link has added a new identity type.

edit/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    if let Some(updated_payload) = term::Editor::new().edit(&payload).unwrap() {
```
FintanH, line 90: Mostly a stylistic thing, but would it not make more sense to have this code in the above match rather than matching on `identity` twice? It just comes at the cost of duplicating the `term::Editor` code.

FintanH (approved):
LGTM

edit/src/lib.rs
cloudhead: Agree that's preferrable to re-matching

edit/src/lib.rs
cloudhead: `anyhow!("unknown identity type for {}", urn)`

edit/src/lib.rs
xphoniex: Made the changes [here](https://github.com/radicle-dev/radicle-cli/compare/ec0b9679b7a5e43fc48691fca49f262e317765bb..872886074161ff6e140725576f4ffffedb30e99b) but I find the one with re-match more readable.

edit/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        .or_else(|| radicle_common::project::cwd().ok().map(|(urn, _)| urn))
+        .ok_or_else(|| anyhow!("Couldn't get URN from either command line or cwd"))?;
+
+    let identity = any::get(&storage, &urn)?
+        .ok_or_else(|| anyhow!("No project or person found for this URN"))?;
+
+    match identity {
+        SomeIdentity::Project(_) => {
+            let payload =
+                serde_json::to_string_pretty(project::verify(&storage, &urn)?.unwrap().payload())?;
```
cloudhead, line 83: What is the `unwrap()` here?

cloudhead, line 83: What is the `unwrap()` here?

xphoniex: Since we've already loaded the `urn` using `any::get` once and we know it exists, we can safely call unwrap on the `Option` here.

cloudhead: myeah... since you're reading from the file-system you actually can't guarantee that. It's too bad there's no `any::verify` or something like that..

xphoniex: `::verify` functions return `Result<Option<VerifiedX>, Error>`.

I assumed if there's a fs failure, we'd get an `Error` from `Result`, no?

cloudhead: Yeah, but if some other process removed the identity in the meantime, you'd get a panic instead of a "not found" error

xphoniex: Ok, fixed https://github.com/radicle-dev/radicle-cli/commit/3ff55c8f057a810425db5a3f3c6d15ae02c9bc79#diff-7f6d0944507bd7f26b50af3f4bce4555849243552f506948e0acbbffe597f4f8R83-R84

xphoniex: > myeah... since you're reading from the file-system you actually can't guarantee that. It's too bad there's no `any::verify` or something like that..

also `any::verify` would be easy to implement. what do you think @FintanH ? I can take of that.

cloudhead (approved):

cloudhead:
Merged in 93a1bfb

edit/src/lib.rs
FintanH: I took a quick look at the identity code, and I'm not sure how easy it would be to do it efficiently. I wouldn't accept a patch that tries to load `project::verify` and if that fails then try `person::verify`. But if you're willing to do dig in and find a way to get `SomeVerifiedIdentity` then I'm happy to review that :)

PR #112, opened 2022-05-24
common: Serialize char keys in Comment::reactions

sebastinez:
This commit introduces the `serialize_reactions` function which is being
used in the `reactions` field in the `Comment` struct.

To be able to convert `Comment` struct to JSON and return them e.g. in
the http-api we need to convert the
`char` into `String` during serialization.

Signed-off-by: Sebastian Martinez <sebastinez@me.com>

### DIFF
--- a/common/src/cobs/shared.rs
+++ b/common/src/cobs/shared.rs
@@ -26,6 +26,7 @@ pub enum ResolveError {
 pub type Discussion = Vec<Comment<Replies>>;
 
 #[derive(Debug, PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct Reaction {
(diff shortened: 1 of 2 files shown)


common/src/cobs/shared.rs
```diff
@@ -134,11 +135,26 @@ pub type Replies = Vec<Comment>;
 pub struct Comment<R = ()> {
     pub author: Author,
     pub body: String,
+    #[serde(serialize_with = "serialize_reactions")]
```
cloudhead, line 138: I think this would be more easily achieved with `#[serde(transparent)]` on the `Reaction` struct, if I'm not mistaken.

sebastinez: Nice good catch, did reduce the PR to just that. 😅

PR #113, opened 2022-05-24
rad-issue: Add `rad issue reaction`

sebastinez:
Allows adding emoji reactions as string to a comment specified by its number,
being 0 the root comment.

**TODO:**
- [x] Improve the way to select a comment to which react, eventually a selectable comment listing

### DIFF
--- a/issue/src/lib.rs
+++ b/issue/src/lib.rs
@@ -19,8 +19,8 @@ pub const HELP: Help = Help {
     usage: r#"
 Usage
 
-    rad issue create [--title <title>] [--description <text>]
-    rad issue remove <id>
+    rad issue new [--title <title>] [--description <text>]
+    rad issue delete <id>
     rad issue comment <id> [--description <text>]
     rad issue react <id> [--emoji <char>]
     rad issue list
@@ -41,8 +41,8 @@ pub struct Metadata {
 pub enum OperationName {
     Create,
     Comment,
-    Reaction,
-    Remove,
+    React,
+    Delete,
     List,
 }
 
@@ -58,12 +58,12 @@ pub enum Operation {
         title: Option<String>,
         description: Option<String>,
     },
-    Remove {
+    Delete {
         id: cobs::issue::IssueId,
     },
-    Reaction {
+    React {
         id: cobs::issue::IssueId,
-        emoji: String,
+        reaction: cobs::Reaction,
     },
     Comment {
         id: cobs::issue::IssueId,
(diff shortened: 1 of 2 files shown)


issue/src/lib.rs
```diff
@@ -38,6 +40,8 @@ pub struct Metadata {
 #[derive(Debug, PartialEq, Eq)]
 pub enum OperationName {
     Create,
+    Comment,
+    Reaction,
```
cloudhead, line 44:
```suggestion
    React,
```

issue/src/lib.rs
```diff
@@ -57,6 +61,14 @@ pub enum Operation {
     Remove {
         id: cobs::issue::IssueId,
     },
+    Reaction {
+        id: cobs::issue::IssueId,
+        emoji: String,
```
cloudhead, line 66: We should use `char` here.

issue/src/lib.rs
```diff
@@ -84,13 +97,20 @@ impl Args for Options {
...
+                Long("description")
+                    if op == Some(OperationName::Create) || op == Some(OperationName::Comment) =>
+                {
                     description = Some(parser.value()?.to_string_lossy().into());
                 }
                 Value(val) if op.is_none() => match val.to_string_lossy().as_ref() {
-                    "c" | "create" => op = Some(OperationName::Create),
-                    "r" | "remove" => op = Some(OperationName::Remove),
+                    "n" | "new" => op = Some(OperationName::Create),
+                    "rm" | "remove" => op = Some(OperationName::Remove),
```
cloudhead, line 110: Let's switch to the new commands `[d]elete` and `[n]ew`

issue/src/lib.rs
```diff
@@ -137,6 +165,17 @@ pub fn run(options: Options) -> anyhow::Result<()> {
         } => {
             issues.create(&project, &title, &description, &[])?;
         }
+        Operation::Reaction { id, emoji } => {
+            if let Some(issue) = issues.get(&project, &id)? {
+                let comment_id = term::comment_select(&issue).unwrap();
```
erak, line 170: Mmh, calling `rad issue react <id> --emoji` panics on an issue that has no comments yet. Is this intended? Otherwise we might want to handle this case more gracefully.

sebastinez: Hmm weird, if there are no comments on an issue it should only allow to select the root comment.. What type of errror you get when it panics?

erak: I'm so sorry! Accidentally tried with an older commit on your branch. Is working as intended.

sebastinez: No problem! Thank you for testing it out!

cloudhead (approved):
Looks great!

PR #122, opened 2022-05-31
rad-issue: Close and re-open issues

sebastinez:
This PR adds the following commands and functionalities

## commands:
- `rad issue state <id> [--close | --open] [--solved]`

Allows the user to change the state of an issue from `open` to `closed, give a `CloseReason` by using the `--solved` flag, and re-open the issue if necessary.

### DIFF
--- a/common/src/cobs/issue.rs
+++ b/common/src/cobs/issue.rs
@@ -61,6 +61,15 @@ pub enum State {
     Closed { reason: CloseReason },
 }
 
+impl State {
+    fn lifecycle_message(self) -> String {
+        match self {
+            State::Open => "Open issue".to_owned(),
+            State::Closed { .. } => "Close issue".to_owned(),
+        }
+    }
+}
+
 impl From<State> for ScalarValue {
     fn from(state: State) -> Self {
         match state {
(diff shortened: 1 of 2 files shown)


sebastinez:
> Might be simpler if you implemented a `issue::lifecycle` function that took a `issue::State`

Okay, yeah that makes sense

issue/src/lib.rs
```diff
@@ -78,6 +85,13 @@ pub struct Options {
     pub op: Operation,
 }
 
+/// State options.
+#[derive(Debug)]
+pub enum StateOptions {
+    Close,
+    Open,
+}
```
cloudhead, line 93: Why not use the issue::State enum?

issue/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub const HELP: Help = Help {
 Usage
 
     rad issue new [--title <title>] [--description <text>]
+    rad issue state <id> [--close | --open] [--solved]
```
cloudhead, line 23: I would collapse it and make them states: `[--closed | --open | --solved]`, so there's no need to do `--close --solved`

common/src/cobs/issue.rs
```diff
@@ -472,6 +466,15 @@ mod events {
...
     };
 
+    impl From<State> for String {
+        fn from(state: State) -> Self {
+            match state {
+                State::Open => "Open issue".to_owned(),
+                State::Closed { .. } => "Close issue".to_owned(),
+            }
+        }
+    }
```
cloudhead, line 476: This will apply to *all* conversions from State to String, so it's not what we want. If you want something like this, I'd implement a private function on `State` called `lifecycle_message`

issue/src/lib.rs
```diff
@@ -97,6 +104,19 @@ impl Args for Options {
                 Long("title") if op == Some(OperationName::Create) => {
                     title = Some(parser.value()?.to_string_lossy().into());
                 }
+                Long("close") if op == Some(OperationName::State) => {
```
cloudhead, line 107: Since we're setting the state, we should call this one `"closed"`

issue/src/lib.rs
cloudhead: Doesn't look like this was solved :)

sebastinez: 😓 sorry yeah will add it asap

sebastinez: Ok has been solved

PR #124, opened 2022-05-31
rad-reward: add error context, move primitives to `common`, add `quickcheck` test

xphoniex:


reward/src/lib.rs
```diff
@@ -192,7 +192,9 @@ pub fn run(options: Options) -> anyhow::Result<()> {
                 Err(_) => return None,
                 Ok(sha1) => sha1,
             };
-            let fp = git::commit_ssh_fingerprint(Path::new("."), &sha1.to_string()).ok()?;
+            let fp = git::commit_ssh_fingerprint(Path::new("."), &sha1.to_string())
+                .map_err(|e| anyhow!("failed to get commit's ssh fingerprint: {}", e))
+                .ok()?;
```
cloudhead, line 198: We should get rid of the `ok()` if we want this error context to actually surface, as this will silence it.

xphoniex: Ah, we actually can't return context info here as we're inside the closure, we need to omit  https://github.com/radicle-dev/radicle-cli/pull/124/commits/67a594e31466ed1a5bbde4a4b9e51f77165aed2c , would that work?

cloudhead: Yeah it's fine.

cloudhead (approved):

PR #127, opened 2022-06-02
Depend on a single tokio feature-set to speed up compilation

adaszko:


Cargo.toml
```diff
@@ -138,7 +137,8 @@ members = [
   "untrack",
   "gov",
   "edit",
-  "reward"
+  "reward",
+  "proof-generator",
```
cloudhead, line 141: Don't want to build this everytime, we aren't using it atm.

cloudhead:
Nice.

cloudhead (approved):

PR #140, opened 2022-06-10
rad-auth: Add warning for when `ssh-agent` is not running

xphoniex:
Running `rad auth` when no ssh-agent exists succeeds and when you run any other `rad` command you'll be asked to re-enter password. This is confusing and a warning would be helpful when running `auth`.

### DIFF
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -266,7 +266,10 @@ pub fn authenticate(
         } else {
             term::success!("Signing key already in ssh-agent");
         }
-    }
+    } else {
+        term::warning("Radicle key won't be added to ssh-agent since it's not running.");
+        term::blank();
+    };
 
     Ok(())
 }


erak:
@cloudhead @xphoniex Shall we re-activate this? I think it's useful :slightly_smiling_face:

xphoniex:
It has saved me a few times tbh, when `rad sync` was not working.

auth/src/lib.rs
```diff
@@ -289,6 +289,8 @@ pub fn authenticate(profiles: &[profile::Profile], options: Options) -> anyhow::
         }
         sock.to_signer(profile)?
     } else {
+        term::warning("Radicle key won't be added to ssh-agent since it's not running. You can run it with `eval $(ssh-agent)`.");
```
cloudhead, line 292: I'm not sure I would include instructions on running ssh-agent, as it might be different based on the OS, eg. I think on macOS you'd use launchd.

xphoniex: it's also terminal-specific as well. e.g. in fish `eval (ssh-agent -c)`.

do you want it entirely omitted? or should I adjust msg per platform? (would be hard to get 100% right)

erak: In order to keep this rather simple, I'd propose to remove the hint entirely and just keep the warning about `ssh-agent` not running.

erak (approved):
LGTM :slightly_smiling_face:

PR #145, opened 2022-06-15
cobs: Add issue and patches count fn

sebastinez:
Signed-off-by: Sebastian Martinez <me@sebastinez.dev>


common/src/cobs/patch.rs
```diff
@@ -476,6 +486,8 @@ pub struct Revision {
     pub merges: Vec<Merge>,
     /// When this revision was created.
     pub timestamp: Timestamp,
+    /// Changeset of this revision.
+    pub changeset: Option<Diff>,
```
cloudhead, line 490: What's going on here? Should this be in another PR? Looks like it's just set to `None`

sebastinez: Yeah seems like the PRs got mixed up, and this leaked into this PR. Will separate it today

cloudhead (approved):

PR #155, opened 2022-06-27
Add release profile to Cargo.toml

xphoniex:
This profile significantly reduces the size of release binaries. As an example `rad-gov` is now only 5.8 MB, down from 20 MB.

We can also use `upx --best --lzma <bin>` and get up to 70% more compression on the binary. Basic testing shows around 200ms added to startup time, which depending on the use case might or might not be worth it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,12 @@ inherits = "dev"
 debug = 0
 incremental = false
 
+[profile.release]
+strip = true
+lto = true
+codegen-units = 1
+panic = "abort"
+
 [features]
 default = ["ethereum"]
 ethereum = ["rad-ens", "rad-account", "rad-help/ethereum", "rad-gov", "rad-reward", "ethers", "futures"]


Cargo.toml
```diff
@@ -11,6 +11,12 @@ inherits = "dev"
 debug = 0
 incremental = false
 
+[profile.release]
```
xphoniex, line 14: Given that the majority of savings comes from `strip`, I decided to leave `opt-level` to the default (which is `3` for release) instead of `"s"` which would've shaved another ~800kb (16%) off `rad-gov`. Thus our binary is still optimized for performance.

cloudhead (dismissed):
:+1:

cloudhead:
Can you include the size gains in the commit message like you shared on discord?

xphoniex:
> Can you include the size efficiency gains in the commit message like you shared on discord?

Added to https://github.com/radicle-dev/radicle-cli/pull/155/commits/61394042ed180bfe161847cf2afee87c0252aa68

cloudhead:
Can you rebase this on master, I fixed the build issue on master.

xphoniex:
> Can you rebase this on master, I fixed the build issue on master.

Done.

Issue #159, opened 2022-06-30
Patch improvements


cloudhead:
While working through the patch workflow, we encounted some issues.

* [ ] `rad-patch` pushes code before patch is confirmed
* [ ] `rad-patch --list` doesn't show branch names
* [ ] `rad-remote rm` has no output if it doesn't remove anything
* [ ] `rad-untrack` should have the option to remove working copy refs/branches/remotes (`on` by default)
* [ ] `rad-track` doesn't tell user that it setup branch and remote
* [ ] `rad-review` doesn't have an option to sync review
* [ ] `rad-patch --list` doesn't have option to sync, so it's out of date unless you `rad sync --fetch` first

slack-coder:
'rad-patch pushes code before patch is confirmed'

I recommend removing git-push confirmation from patch creation.  Instead, either check and bail out if the code is uploaded with a flag to auto-push the code, or allow patch creation without requiring the code in radicle's local storage/monorepo.

The later option would be cool, but its harder if its even possible.  Users would have more freedom to use radicle patches in their own way, with any git hosting system.

I prefer the former option for practicality.  The action requires the code to be in local storage/monorepo at least ahead of any conceptual sanity checks; its a soft constraint.  Therefor its somewhat implicit in patch creation.

PR #163, opened 2022-07-04
remote: report error when removing unknown peer

slack-coder:
Fail with a message instead of silently failing when an unknown remote peer is removed with the 'rm' sub-command.

Issue #159

### DIFF
--- a/remote/src/lib.rs
+++ b/remote/src/lib.rs
@@ -4,7 +4,9 @@ use std::ffi::OsString;
 use anyhow::anyhow;
 use anyhow::Context as _;
 
+use librad::git::storage::Storage;
 use librad::git::tracking;
+use librad::git::Urn;
 use librad::PeerId;
 
 use radicle_common::args::{Args, Error, Help};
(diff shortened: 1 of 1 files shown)


cloudhead (commented):
This only seems to solve the problem when a name is passed. When it's a peer-id, we still fail quietly..

remote/src/lib.rs
```diff
@@ -171,21 +171,33 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+                return Ok(())
+            }
+
+            // Delete by peer name.
+            let mut x = None;
+            for (name, peer) in git::remotes(&repo)? {
+                if let Some(person) = project::person(&storage, urn.clone(), &peer)? {
+                    if person.subject().name.to_string() == remote {
+                        x = Some((name, &remote));
+                        break;
```
cloudhead, line 184: Why was this block of code moved out of the if/else above?

slack-coder: It was a vain attempt at reducing the code's indentation level.  In the end, I preserve the if/else structure in the proposed change.

remote/src/lib.rs
```diff
@@ -159,33 +159,40 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
             );
         }
         Operation::Remove { remote } => {
+            let mut remote_found = None;
```
cloudhead, line 162: This works, but what if we moved this if/else into a function? It would simplify the control flow a lot, ie. we could do `match find_remote(...)? { ... } `

PR #168, opened 2022-07-07
Support message flags for patch creation

slack-coder:
Use '--message' to set the patches title and description on creation to reduce required human intervention.  The description field is also made optional.

### DIFF
--- a/scripts/contributor-workflow-setup.sh
+++ b/scripts/contributor-workflow-setup.sh
@@ -94,7 +94,7 @@ git commit -m "Update README" --no-gpg-sign
 # Push commit to monorepo
 rad push
 # Create patch
-rad patch --sync
+rad patch --sync --message "Update README" --message "Reflect the recent positive news"
 # Sync identity
 rad sync --self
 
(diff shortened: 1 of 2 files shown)


scripts/contributor-workflow-setup.sh
```diff
@@ -94,7 +94,7 @@ git commit -m "Update README" --no-gpg-sign
 # Push commit to monorepo
 rad push
 # Create patch
-rad patch --sync
+rad patch --sync --message 'title' --message "description"
```
cloudhead, line 97:
```suggestion
rad patch --sync --message 'Update README' --message "No description..."
```

cloudhead: Will be useful to have a `--no-edit` flag next, so that it just re-uses the commit message without asking for input.

cloudhead (commented):
Looks good overall, just a couple of minor things.

scripts/contributor-workflow-setup.sh
```diff
@@ -94,7 +94,7 @@ git commit -m "Update README" --no-gpg-sign
 # Push commit to monorepo
 rad push
 # Create patch
-rad patch --sync
+rad patch --sync --message "Update README" --message "Reflect the recent positive news"
```
cloudhead, line 97: :laughing: :+1:

PR #171, opened 2022-07-08
terminal-tui: library w/ tui-realm

erak:

PR #172, opened 2022-07-08
issue: TUI

erak:
### Demo

https://user-images.githubusercontent.com/20012009/178080096-8720e0de-5149-4385-801d-7170b0187f05.mp4

### Description

Requires https://github.com/radicle-dev/radicle-cli/pull/171


issue/src/tui/components.rs
```diff
@@ -192,3 +208,105 @@ impl Component<Message, NoUserEvent> for IssueList {
         }
     }
 }
+
+pub struct CommentList<R> {
+    props: Props,
+    comments: ListState<Comment<R>>,
+}
```
cloudhead, line 221:
```rust
// before
SpecificListComponent<R> {
   state: SharedState<Specific<R>>
}
// after
GenericList<ChildComponent> {
   children: Vec<ChildComponent>,
   state: SpecificState
}
```

erak: I'm not sure this will work since every component must provide a component-specific implementation of `Component` in order to receive events: https://github.com/radicle-dev/radicle-cli/pull/172/files#diff-dcbc1cad867aa6da05adc9efaa047bda13bfcdf930b86e9b17bcb8544163eb1bR365. 

They usually call `self.perform()` which would need to generate component-specific command results like "index changed in issue list". Might need to experiment with a generic implementation and see if that would fit in.

PR #174, opened 2022-07-11
patch: list patches using the monorepo

slack-coder:
Remove the dependency on using the working copy to list patches.

The new output looks as below, with working copy information
removed.

```
title R0 ab0336c
└── hnrkqehrics opened by scooby 3 days ago
    ✓ accepted by cloudhead  delegate  (you) 3 days ago
```


slack-coder:
@cloudhead what's your opinion about the output without the git history information from the working copy?

cloudhead:
Why was the information removed and not simply fetched from the monorepo?

cloudhead:
Alternatively we could check if the working copy (if any) has the information, and if not we `git fetch` first.

cloudhead (approved):

PR #175, opened 2022-07-12
patch: improve clarity of list source code

slack-coder:
Lift creating the message for git sync status of a patch into its own function.

### DIFF
--- a/patch/src/lib.rs
+++ b/patch/src/lib.rs
@@ -532,6 +532,28 @@ fn create(
     Ok(())
 }
 
+/// Create a human friendly message about git's sync status.
+fn pretty_sync_status(
+    repo: &git::Repository,
+    revision_oid: git::Oid,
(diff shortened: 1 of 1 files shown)


patch/src/lib.rs
```diff
@@ -532,6 +532,29 @@ fn create(
...
+fn pretty_sync_status(
+    repo: &git::Repository,
+    revision_oid: git::Oid,
+    head: &git::Reference,
+) -> anyhow::Result<String> {
+    let target = head.target();
+    if target == None {
+        return Ok(String::default());
+    }
+    let head_oid = target.unwrap();
```
cloudhead, line 545: No need to unwrap here if you use `if let`

slack-coder: Noted, thanks :)

slack-coder:
changes included in #174 174

PR #181, opened 2022-07-14
patch: Sync before listing patch info

slack-coder:
Make it easier for the user to list recent patch information by supporting the sync option by the patch --list command.


patch/src/lib.rs
```diff
@@ -189,7 +190,22 @@ fn list(
     repo: &git::Repository,
     profile: &Profile,
     project: &project::Metadata,
+    options: Options,
 ) -> anyhow::Result<()> {
+    // function uses information changed by sync
```
cloudhead, line 195: I don't think the comment is necesssary, it confused me more than anything :stuck_out_tongue_closed_eyes:

cloudhead:
Rebased and merged in 77017adfbf74a822424e4f1b94ffaf958bbe428b

PR #182, opened 2022-07-14
track: dry up track syncing

slack-coder:
Reduce code by pushing conditional logic to determining the sync's seed argument.


cloudhead (approved):
lgtm

PR #185, opened 2022-07-15
Make '--message' consistent across commands

slack-coder:
Update '--comment' and '--review' to use the '--message' option the same way as patch where additional '--message' arguments are appended together.

### DIFF
--- a/comment/src/lib.rs
+++ b/comment/src/lib.rs
@@ -10,6 +10,7 @@ use radicle_common::{
     keys, project,
 };
 use radicle_terminal as term;
+use radicle_terminal::patch::Comment;
 
 pub const HELP: Help = Help {
     name: "comment",
@@ -31,7 +32,7 @@ Options
 #[derive(Debug)]
 pub struct Options {
     pub id: cobs::Identifier,
-    pub message: Option<String>,
+    pub message: Comment,
     pub reply_index: Option<CommentId>,
 }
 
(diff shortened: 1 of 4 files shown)


slack-coder:
@cloudhead now that patch::Comment is being used by other commands, maybe we should move it?  Where's a good place for it?

terminal/src/patch.rs
```diff
@@ -32,6 +32,15 @@ impl Comment {
 
         comment.to_owned()
     }
+
+    pub fn append(&self, arg: &str) -> Comment {
```
cloudhead, line 36: Since you're returning a new comment here, I'd have this function move `self` instead of taking it by reference. This will also avoid others using this function like `comment.append("...");` and not understanding why nothing was appended.

slack-coder: I simplified the code by making append mutate self in place of returning a new copy.

comment/src/lib.rs
```diff
@@ -22,7 +23,7 @@ Usage
 
 Options
 
-    -m, --message               Comment message
+    -m, --message [<string>]    Provide a comment with the review (default: prompt)
```
cloudhead, line 26: Note that this is the `comment` command, so it's not necessarily used with a review, hence I'd leave the description as-is.

terminal/src/patch.rs
```diff
@@ -32,6 +32,14 @@ impl Comment {
 
         comment.to_owned()
     }
+
+    pub fn append(&mut self, arg: &str) {
+        if let Comment::Text(v) = self {
+            *self = Comment::Text(v.to_owned() + "\n\n" + arg)
```
cloudhead, line 38: You could just do `v.extend(["\n\n", arg]);` here

cloudhead:
Looking good.

PR #188, opened 2022-07-25
untrack: cleanup the peer's rad remote

slack-coder:
Remove the projects rad remote for the peer being untracked.  This is to aid cleaning up and undoing the effects of the 'track' command.

When run from outside a project context via external commands, like 'rad-rm', the cleanup step is skipped.

#159

### DIFF
--- a/terminal/src/remote.rs
+++ b/terminal/src/remote.rs
@@ -1,0 +1,64 @@
+use std::collections::HashSet;
+
+use librad::git::storage::Storage;
+use librad::git::Urn;
+
+use radicle_common::{git, project};
+
+use crate as term;
+
+pub fn list(storage: &Storage, repo: &git::Repository, urn: &Urn) -> anyhow::Result<()> {
+    let mut table = term::Table::default();
+    let proj = project::get(&storage, urn)?
+        .ok_or_else(|| anyhow::anyhow!("project {} not found on local device", urn))?;
+    let mut peers = HashSet::new();
+
+    for (_, peer) in git::remotes(repo)? {
+        if !peers.insert(peer) {
+            // Don't show duplicate peers.
+            continue;
+        }
+
+        let delegate = if proj.remotes.contains(&peer) {
+            term::format::badge_primary("delegate")
+        } else {
+            String::new()
+        };
+
+        if let Some(person) = project::person(&storage, urn.clone(), &peer)? {
+            table.push([
+                term::format::bold(person.subject().name.to_string()),
+                term::format::tertiary(peer),
+                delegate,
+            ]);
+        } else {
+            table.push([String::new(), term::format::tertiary(peer), delegate]);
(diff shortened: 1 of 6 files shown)


slack-coder:
These changes also support untracking a peer by their name.

slack-coder:
@cloudhead what's your opinion on the argument naming '--cleanup'?  I am unsure about it.

cloudhead:
Yeah, also not sure. How about we do it the other way around and have a `--keep-remote` argument that is off by default? I'd also be ok with not having any flag and always deleting the local config until users ask for it.

terminal/src/remote.rs
```diff
@@ -0,0 +1,91 @@
...
+    // by person's name
+    for (name, peer) in git::remotes(repo)? {
+        if let Some(person) = project::person(&storage, urn.clone(), &peer)? {
+            if person.subject().name.to_string() == *name {
+                return Ok(Some(name));
+            }
+        }
+    }
+    Ok(None)
+}
```
cloudhead, line 35: Can this function maybe go in `common/project`?

untrack/src/lib.rs
```diff
@@ -76,21 +72,51 @@ impl Args for Options {
...
+        } else {
+            let project = project::get(&storage, urn)?
+                .ok_or_else(|| anyhow!("project {} not found in local storage", &urn))?;
+
+            if let Some(v) = get_peer_id(&project, &storage, &peer_str)? {
+                v
+            } else {
+                anyhow::bail!("invalid <peer-id> '{}'", peer_str)
+            }
+        };
```
cloudhead, line 118: Ok so this allows untracking a peer by name?

untrack/src/lib.rs
```diff
@@ -76,21 +72,51 @@ impl Args for Options {
...
-    // TODO: Remove tracking branch
+fn get_peer_id(
+    project: &project::Metadata,
+    storage: &Storage,
+    name: &String,
+) -> anyhow::Result<Option<PeerId>> {
+    let mut found = None;
+    for (id, peer) in project::tracked(project, storage)? {
+        if peer.name() == *name {
+            found = Some(id);
```
cloudhead, line 90: Don't we want to return here? Ofc. there's also the issue of multiple users with the same name, but can tackle that later..

terminal/src/remote.rs
slack-coder: done

untrack/src/lib.rs
slack-coder: Yes, I should have referenced this in the commit message

cloudhead:
Hey I tried to squash this but didn't manage, can you rebase and squash?

slack-coder:
> Hey I tried to squash this but didn't manage, can you rebase and squash?

All good

PR #190, opened 2022-07-27
Show branch names when listing patches

slack-coder:
Show branch names when listing patches, or nothing when none are present.  This gives the following output:

```
Update README hnrkkqa8n56 R0 f092fb4(change-readme,rad/change-readme) ahead 1, behind 0
└── opened by hnrkd471zrxzjgct4gestio35ai7wnhq67yeo 30 minutes ago
```

### DIFF
--- a/patch/src/lib.rs
+++ b/patch/src/lib.rs
@@ -173,7 +173,7 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
         .ok_or_else(|| anyhow!("couldn't load project {} from local state", urn))?;
 
     if options.list {
-        list(&storage, &profile, &project, options)?;
+        list(&storage, Some(repo), &profile, &project, options)?;
     } else {
         create(&storage, &profile, &project, &repo, options)?;
     }
@@ -183,6 +183,7 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
 
 fn list(
     storage: &Storage,
+    repo: Option<git::Repository>,
     profile: &Profile,
     project: &project::Metadata,
     options: Options,
@@ -227,7 +228,7 @@ fn list(
         for (id, patch) in &mut own {
             term::blank();
 
-            print(&cobs.whoami, id, patch, project, &monorepo, storage)?;
+            print(&cobs.whoami, id, patch, project, &monorepo, &repo, storage)?;
         }
     }
     term::blank();
@@ -240,7 +241,7 @@ fn list(
         for (id, patch) in &mut other {
             term::blank();
 
-            print(&cobs.whoami, id, patch, project, &monorepo, storage)?;
+            print(&cobs.whoami, id, patch, project, &monorepo, &repo, storage)?;
         }
     }
     term::blank();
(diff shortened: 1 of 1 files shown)


scripts/contributor-workflow-setup.sh
```diff
@@ -110,6 +110,7 @@ CONTRIBUTOR_PEER=$(rad self --peer)
 cd acme
 
 # Create change
+git checkout -b update-readme
```
cloudhead, line 113: Does that not break anything?

patch/src/lib.rs
```diff
@@ -555,13 +556,40 @@ fn pretty_sync_status(
...
+        for r in repo.references()?.flatten() {
+            if let (Some(oid), Some(name)) = (&r.target(), &r.shorthand()) {
+                if oid == revision_oid {
+                    branches.push(name.to_string());
+                };
+            };
+        }
+    };
+    if !branches.is_empty() {
+        str = format!("{}({})", str, &branches.join(","));
```
cloudhead, line 579:
```suggestion
        str = format!("{} ({})", str, &branches.join(", "));
```

patch/src/lib.rs
```diff
@@ -555,13 +556,40 @@ fn pretty_sync_status(
     Ok(format!("ahead {}, behind {}", ahead, behind))
 }
 
+/// Make a human friendly string for commit version information.
+///
+/// For example 'oid(branch1[,branch2])'.
```
cloudhead, line 561:
```suggestion
/// For example '<oid> (branch1, branch2)'.
```

patch/src/lib.rs
```diff
@@ -555,13 +556,40 @@ fn pretty_sync_status(
...
+fn pretty_commit_version(
+    revision_oid: &git::Oid,
+    repo: &Option<git::Repository>,
+) -> anyhow::Result<String> {
+    let mut str = common::fmt::oid(revision_oid);
+
+    let mut branches: Vec<String> = vec![];
```
cloudhead, line 568: I'd group the `let` declarations

cloudhead:
Looks good overall, just a few minor things.

scripts/contributor-workflow-setup.sh
slack-coder: Nope.  Yet I updated it to checkout master afterwards to avoid any future problems.

cloudhead: For me it's broken:
```
fatal: commit hashes do not match: 1d09e3c1591bb81d63a78bc86a1a89cb38c2681d vs. c28ebbf6fe1ef53e4908124369f1a531e3f722f6
```

cloudhead: What was the reason for changing it?

slack-coder: My bad, I didn't notice the error at the end.

The rationale is it being the common workflow to use a feature branch.  The catalyst was the change in this PR to show the branch.

slack-coder: As of now, it appears problematic to do this, so I've reverted the commits re: the feature branch.

patch/src/lib.rs
```diff
@@ -555,13 +556,40 @@ fn pretty_sync_status(
...
+) -> anyhow::Result<String> {
+    let mut str = common::fmt::oid(revision_oid);
+    let mut branches: Vec<String> = vec![];
+
+    if let Some(repo) = repo {
+        for r in repo.references()?.flatten() {
+            if let (Some(oid), Some(name)) = (&r.target(), &r.shorthand()) {
+                if oid == revision_oid {
+                    branches.push(name.to_string());
+                };
```
cloudhead, line 574: Ok one thing I noticed here, is I'm getting things other than branches match, eg. remote refs that aren't associated to local branches. So I think we need an additional check, maybe by trying to fetch the branch or something?

cloudhead: Going to merge this already, can tackle it later.

PR #194, opened 2022-08-01
Slackcoder/track show peer branches

slack-coder:
Show all peer branches when listing peers through rad-track.

References #159 for 'rad-track --local should show non-default branches too'

### DIFF
--- a/track/src/lib.rs
+++ b/track/src/lib.rs
@@ -15,7 +15,7 @@ use radicle_common::nonempty::NonEmpty;
 use radicle_common::project::PeerInfo;
 use radicle_common::tokio;
 use radicle_common::Url;
-use radicle_common::{git, keys, project, seed, sync};
+use radicle_common::{git, keys, project, seed, sync, Urn};
 use radicle_terminal as term;
 
 mod options;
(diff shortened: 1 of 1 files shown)


slack-coder:
@cloudhead should I move get_peer_branches to common:git?  It also looks like get_remote_head isnt used anymore.

cloudhead:
I think it's fine where it is for now. I'd like to keep the `git` module free from dependencies on librad.

track/src/lib.rs
```diff
@@ -283,32 +283,57 @@ pub fn show(
     Ok(())
 }
 
+// return the Peer's branches with their Oid's by parsing references in the storage's git
+// repository.
```
cloudhead, line 287: Docstrings start with `///` and a capital

track/src/lib.rs
```diff
@@ -283,32 +283,57 @@ pub fn show(
     Ok(())
 }
 
+/// return the Peer's branches with their Oid's by parsing references in the storage's git
```
cloudhead, line 286:
```suggestion
/// Teturn the Peer's branches with their Oid's by parsing references in the storage's git
```

PR #195, opened 2022-08-02
Reduce import complexity for git mod

slack-coder:
Use common::git in place of the git2 module in many cases.  Git2 is re-exported by common::git as the project's git library.

### DIFF
--- a/common/src/project.rs
+++ b/common/src/project.rs
@@ -7,7 +7,6 @@ use std::path::PathBuf;
 
 use anyhow::{anyhow, Result};
 use either::Either;
-use git2::Repository;
 use serde::{Deserialize, Serialize};
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


common/src/project.rs
```diff
@@ -406,14 +405,14 @@ pub fn get_remote_head<S>(
...
+) -> anyhow::Result<Option<git::Oid>>
 where
     S: AsRef<ReadOnly>,
 {
     // Open the monorepo.
-    let repo = git2::Repository::open_bare(storage.as_ref().path())?;
+    let repo = git::Repository::open_bare(storage.as_ref().path())?;
 
-    // Nb. `git2` doesn't handle namespaces properly, so we specify it manually.
+    // Nb. `git` doesn't handle namespaces properly, so we specify it manually.
```
cloudhead, line 415: This is referring to the `git2` crate, so maybe leave it as-is

PR #197, opened 2022-08-03
Removed unused dependencies from project crates

slack-coder:


cloudhead:
Some of these are actually needed by the `ethereum` feature, try to run `cargo check --all-features`

Issue #196, opened 2022-08-02
rad-gov crate not compiling


slack-coder:
The rad-gov create fails to build with the following error.

```
error[E0433]: failed to resolve: could not find `ethereum` in `term`
   --> gov/src/lib.rs:131:48
    |
131 |     let (wallet, provider) = rt.block_on(term::ethereum::get_wallet(signer_opts, provider))?;
    |                                                ^^^^^^^^ could not find `ethereum` in `term`

error[E0433]: failed to resolve: could not find `ethereum` in `term`
   --> gov/src/lib.rs:159:11
    |
159 |     term::ethereum::transaction(call).await?;
    |           ^^^^^^^^ could not find `ethereum` in `term`

error[E0433]: failed to resolve: could not find `ethereum` in `term`
   --> gov/src/lib.rs:241:11
    |
241 |     term::ethereum::transaction(call).await?;
    |           ^^^^^^^^ could not find `ethereum` in `term`

error[E0433]: failed to resolve: could not find `ethereum` in `term`
   --> gov/src/lib.rs:252:11
    |
252 |     term::ethereum::transaction(call).await?;
    |           ^^^^^^^^ could not find `ethereum` in `term`

error[E0433]: failed to resolve: could not find `ethereum` in `term`
   --> gov/src/lib.rs:282:15
    |
282 |         term::ethereum::transaction(call).await?;
    |               ^^^^^^^^ could not find `ethereum` in `term`

```

PR #198, opened 2022-08-03
gov: fix compile time error due to dep error

slack-coder:
Add the 'ethereum' feature to radicle-terminal to fix a compile time
error.

Fixes issue #196


cloudhead:
I can't reproduce that issue, maybe you're trying to build without `ethereum`?

image

slack-coder:
I was yes, I overlooked that part.  I still recommend including this change.  It is a compile time error still, and the feature is used for the other dependency in the package.

PR #200, opened 2022-08-04
Move all sub-commands to `rad` binary

cloudhead:
* Reduce installation size
* Reduce build times

Signed-off-by: Alexis Sellier <self@cloudhead.io>


slack-coder:
I notice 'lexopt::Parser::from_env()' has been replaced with 'lexopt::Parser::from_args(args)' in many places.

Does this bring changes to how the commands operate?

cloudhead:
> I notice 'lexopt::Parser::from_env()' has been replaced with 'lexopt::Parser::from_args(args)' in many places.
> 
> Does this bring changes to how the commands operate?

Those were bugs basically, since the args were never passed in.

PR #203, opened 2022-08-05
git-remote-rad: import into own crate

slack-coder:
Better bring in git-remote-rad into the code base by following command conventions.
 
- Import a modified term::run_command into git-remote-rad
- Expose methods as a library
- Expose functionality as a command line tool


rad/src/bin/git-remote-rad.rs
```diff
@@ -0,0 +1,52 @@
+use git_remote_rad::{run, Options};
+use radicle_common::args::Args;
+use radicle_common::profile;
+
```
cloudhead, line 4: I don't think this file should be inside `rad`, since it's not the name of the command.

cloudhead: I would expect it to be in the `git-remote-rad` package I guess

README.md
```diff
@@ -6,13 +6,15 @@
...
+You can install the Radicle CLI and supporting tools by running the
+following command from inside this repository:
 
+    cargo install --force --locked --path ./git-remote-rad
     cargo install --force --locked --path .
 
 Or directly from our seed node:
 
+    cargo install --force --locked --git https://seed.alt-clients.radicle.xyz/radicle-cli.git git-remote-rad
     cargo install --force --locked --git https://seed.alt-clients.radicle.xyz/radicle-cli.git radicle-cli
```
cloudhead, line 18: This can be combined into one command, same above

git-remote-rad/Cargo.toml
```diff
@@ -7,14 +7,15 @@ license = "GPL-3.0-or-later"
 description = "Git remote helper for radicle"
 
 [dependencies]
-radicle-cli = { path = "../" }
 async-trait = { version = "0.1" }
 log = { version = "0.4" }
 lexopt = { version = "0.2" }
+dialoguer = "0.10.0"
```
cloudhead, line 13: Why is this needed?

PR #204, opened 2022-08-08
README: correct install instructions

slack-coder:
Sync the README's install instructions with the recent migration to use a the 'rad' crate for commands.  Use a relative path to be clear about which directory to run the command from.


slack-coder:
README is being corrected in upcoming changes

PR #206, opened 2022-08-09
contributor-workflow: use long command arguments

slack-coder:
Improve the scripts clarity by using long arguments for all commands.


scripts/contributor-workflow-setup.sh
```diff
@@ -95,9 +95,9 @@ PROJECT=$(rad inspect)
 
 ####################
 banner "CONTRIBUTOR"
-####################
+###################
```
cloudhead, line 98: ?

slack-coder: My thoughts too.  Fixed.

PR #217, opened 2022-08-16
checkout: improve clarity

slack-coder:
- reduce nesting where possible
 - better express peer selection through vector pattern matching to help visualize how vector length maps to the result.


erak (approved):
Great improvements! LGTM

Issue #189, opened 2022-07-25
rad-remote fails after an untrack --all


slack-coder:
The expected behavior is to either list the remaining remotes, or none if they are removed.  

```
▒ rad untrack --all
ok Tracking relationships for rad:git:hnrkrjq98dyjuson9dzuu8utjawenj7x3jjsy removed

▒ rad remote
== Command failed project rad:git:hnrkrjq98dyjuson9dzuu8utjawenj7x3jjsy not found on local device
```
The root of the error is 'rad untrack --all' unexpectedly deletes the project in the monorepo in addition to untracking peers.  This is what 'rad rm' is for.

cloudhead:
Aha, yes, `rad rm` basically calls `rad untrack --all`. We don't control this behavior. I think it *kind of* makes sense, because of you aren't tracking anyone, then there's no use for having the project anymore, though you could argue that you just never want it to be updated?

One option is to just remove the `--all` option from `rad-untrack`, as I agree it's unexpected.

PR #222, opened 2022-08-19
untrack: remove '--all' option

slack-coder:
Reduce redundancy as 'rad untrack --all' is synonymous with 'rad rm'.

Related to issue #189

### DIFF
--- a/rm/src/lib.rs
+++ b/rm/src/lib.rs
@@ -5,6 +5,7 @@ use std::str::FromStr;
 
 use anyhow::anyhow;
 
+use librad::git::tracking::git::tracking;
 use librad::git::Urn;
 use librad::PeerId;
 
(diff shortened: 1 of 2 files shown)


rm/src/lib.rs
```diff
@@ -84,16 +86,31 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
         .join("namespaces")
         .join(options.urn.encode_id());
 
-    if !options.prompt
+    if options.prompt
```
erak, line 89: Wouldn't this lead to the project being removed when the user does not confirm (presses `n`)?

erak: Sorry, I fear the `fixup!` commit is also not doing the right thing. So right now, if the user confirms the removal, nothing happens. Is there a specific reason you changed the logic here?

slack-coder: First of all, thanks for spotting this!

Yes the fixup was wrong as well.  The logic should remain the same, trying to improve the code by isolating the user confirmation to a few lines.

rm/src/lib.rs
```diff
@@ -84,16 +86,31 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
         ))
     {
-        rad_untrack::execute(&options.urn, rad_untrack::Options { peer: None }, &profile)?;
-        fs::remove_dir_all(namespace)?;
-        term::success!("Successfully removed project {}", options.urn);
+        return Ok(());
     }
 
+    // A project is only considered to exist if it is tracked one or more times.
+    tracking::untrack_all(
```
erak, line 99: Mmh, is this equivalent to calling `rad_untrack::execute` with `peer == None`?

slack-coder: Is there something wrong with the implementation?  This is part of the intent of the PR.

erak: As far as I understand the comments in https://github.com/radicle-dev/radicle-cli/issues/189, this PR should only mostly remove the `--all` option from `rad-untrack`. And thus wondering, what the reason to replace `rad_untrack::execute()` with `tracking::untrack_all()` was. Just don't know if is supposed to behave the same or if's different for the very same reason that I don't fully get yet :wink:

slack-coder: It should be ;)  calling rad_untrack::execute with peer = None is the internal equivalent of 'rad untrack --all'.  I was confused whether a problem was spotted or if it was curiosity.

Check the changes at https://github.com/radicle-dev/radicle-cli/pull/222/files#diff-48f638317ddb6a247ea94a4bca3e12f78ade01dc2233a6e818e2fef066abd524L109-L116

erak: Thanks for the explanation!

erak (dismissed):
LGTM

cloudhead:
Needs a rebase since the `rm` changes have been merged.

slack-coder:
@erak @cloudhead the code was re-implemented due to the amount of recent changes.

There is an additional change to [report an error instead of failing](https://github.com/radicle-dev/radicle-cli/pull/222/commits/ca8838389c24dd75d5f9c7cfe7ffe4719cb6f90e).  There is a bug in the current code base which triggers this and is not trivial to remove.

rm/src/lib.rs
```diff
@@ -122,16 +124,38 @@ pub fn run(options: Options, ctx: impl term::Context) -> anyhow::Result<()> {
...
+            )?;
+            fs::remove_dir_all(namespace)?;
+
+            if let Ok((cwd_urn, cwd_repo)) = project::cwd() {
+                if cwd_urn == *urn {
+                    for p in all_untracked.untracked.flatten() {
+                        let err =
+                            term::remote::remove(&p.remote.to_string(), &storage, &cwd_repo, urn);
+                        if let Err(err) = err {
+                            eprintln!("unable to remove git remote '{}': {}", p.remote, err);
```
erak, line 152: Is this expected to fail often? And is the whole removal operation considered errornous if only one of the remotes can't be removed? If so, we could return the error here already such that it bubbles up correctly. Otherwise I'd suggest to collect the remotes that could not be removed and show an error / warning listing these.

slack-coder:
@erak rebased and addressed the last comment in https://github.com/radicle-dev/radicle-cli/pull/222/commits/f0bc255d342109612ab497c795dc43237af61797.

The failure to remove a subset of remotes may be different for each one, and it should not happen often.  Reporting them per line should be ok.

PR #234, opened 2022-08-26
rad-rm: Implement profile removal

### LINKED ISSUE #220: Support identity removal
### Motivation

It came up a few times that users may want to remove identities (because of e.g. lost password)

### Specification

```
Usage

    rad rm <urn | profile-id> [<option>...]

Options

    --no-confirm    Do not ask for confirmation before removal
    --no-password   Bypass password prompt and do not read env var
    --help          Print help
```

Perhaps it would be good to require a password (either prompt or env var e.g. RAD_PASSWORD as discussed on https://github.com/radicle-dev/radicle-cli/issues/213). So if the password is lost, --no-password must be given.

erak:
Closes #220.

### DIFF
--- a/rm/src/lib.rs
+++ b/rm/src/lib.rs
@@ -24,8 +24,9 @@ Usage
 Options
 
     --no-confirm        Do not ask for confirmation before removal
-    --no-passphrase     Bypass password prompt and do not read environment
-                        variable
+    --no-passphrase     If profile id given, bypass passphrase prompt and
+                        do not read environment variable `RAD_PASSPHRASE`
+                        (default: false)
     --help              Print help
 "#,
 };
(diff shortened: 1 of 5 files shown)


erak:
Rebased and added `--stdin` similar to https://github.com/radicle-dev/radicle-cli/pull/237.

rm/src/lib.rs
```diff
@@ -17,82 +19,157 @@ pub const HELP: Help = Help {
...
+        let mut stdin = false;
 
         while let Some(arg) = parser.next()? {
             match arg {
-                Short('i') => {
-                    prompt = true;
+                Long("no-confirm") => {
+                    confirm = false;
+                }
+                Long("no-password") => {
```
cloudhead, line 76: Should be `no-passphrase`

rm/src/lib.rs
```diff
@@ -17,82 +19,157 @@ pub const HELP: Help = Help {
...
+                    require_secret = false;
```
cloudhead, line 77: I find it easier to use the same names as the options for these vars, that way it's clear what they represent, so just `passphrase = false`

rm/src/lib.rs
```diff
@@ -17,82 +19,157 @@ pub const HELP: Help = Help {
...
+                if options.require_secret {
+                    let passphrase = term::read_passphrase(options.stdin, false)?;
+                    if keys::load_secret_key(&profile, passphrase).is_ok() {
+                        profile::remove(&profile)?;
+                    } else {
+                        anyhow::bail!(format!("Invalid passphrase supplied."));
+                    }
+                } else {
+                    profile::remove(&profile)?;
+                }
```
cloudhead, line 163: You can simplify this by exiting if `load_secret_key.is_err()`

erak:
Updated :slightly_smiling_face:

cloudhead (dismissed):

PR #235, opened 2022-08-26
Tell homebrew to auto update the Radicle tap

adaszko:
As per [Homebrew docs](https://docs.brew.sh/Taps):
> Any location and any protocol that Git can handle is fine, although non-GitHub taps require running brew tap --force-auto-update <user/repo> to enable automatic updating.


erak (approved):
Looks good! Great you spotted this :+1:

Issue #213, opened 2022-08-12
[FEATURE] Add environment variable support for `rad auth` 


JoshuaTrujillo15:
## Overview

The `rad auth` command requires authentication via an interactive terminal. Using the `--init` option allows `--passphrase` and `--name`, but passing such variables into the command line is considered insecure (as the CLI notifies you).

Using default environment variables such as `RAD_PASSPHRASE` as an alternative to command line arguments and interactive prompts would enable more automation workflows.

## Requested Feature

Add the following environment variables to the `rad auth` crate.

- `RAD_PASSPHRASE`
- `RAD_NAME` (for initializing new profiles)
- `RAD_PROFILE` (for authenticating existing profiles)

pinkforest:
thx :)

cloudhead:
`RAD_PASSPHRASE` and `RAD_PROFILE` makes sense to me, but I don't understand the usecase of `RAD_NAME`?

JoshuaTrujillo15:
`RAD_NAME` is mostly for the sake of completeness.

Name is used when creating an account, which isn't likely to be part of much (if any) automation, but it is a flag (`--name`) when initializing a new profile, so it made sense to just have an env option.

Though I'm not hard-committed to this one, mostly just `RAD_PASSPHRASE` and `RAD_PROFILE`.

cloudhead:
Cool, yeah I'd like to keep env vars to a minimum, so would stick with the two most useful ones.

gsaslis:
+1 this would also help with profile management on the IDE, where we want to script the call to `rad auth`

erak:
I'll look into this.

PR #237, opened 2022-08-30
auth: Read passphrase from env var or stdin

erak:
Closes https://github.com/radicle-dev/radicle-cli/issues/213.

### DIFF
--- a/auth/src/lib.rs
+++ b/auth/src/lib.rs
@@ -20,8 +20,9 @@ Usage
 
     rad auth [--init | --active] [<options>...] [<profile>]
 
-    A passphrase may be given via the environment variable `RAD_PASSPHRASE`.
-    Using this disables the passphrase prompt.
+    A passphrase may be given via the environment variable `RAD_PASSPHRASE` or
+    via the standard input stream if `--stdin` is used. Using one of these
+    methods disables the passphrase prompt.
 
     If `--init` is used, a name may be given via the `--name` option. Using
     this disables the input prompt.
@@ -30,6 +31,7 @@ Options
 
     --init                  Initialize a new identity
     --active                Authenticate with the currently active profile
+    --stdin                 Read passphrase from stdin (default: false)
     --name <name>           Use given name (default: none)
     --help                  Print help
 "#,
@@ -39,6 +41,7 @@ Options
 pub struct Options {
     pub init: bool,
     pub active: bool,
+    pub stdin: bool,
     pub name: Option<String>,
     pub profile: Option<ProfileId>,
 }
@@ -49,6 +52,7 @@ impl Args for Options {
 
         let mut init = false;
         let mut active = false;
+        let mut stdin = false;
         let mut name = None;
         let mut profile = None;
         let mut parser = lexopt::Parser::from_args(args);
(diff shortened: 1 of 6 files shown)


auth/src/lib.rs
```diff
@@ -274,17 +260,13 @@ pub fn authenticate(
...
-            } else {
-                let mut input: Zeroizing<String> = Zeroizing::new(Default::default());
-                std::io::stdin().read_line(&mut input)?;
-                SecUtf8::from(input.trim_end())
-            };
-            let pass = keys::pwhash(secret_input);
+            let passphrase = term::read_passphrase(options.stdin, false)?;
+            let secret = keys::pwhash(passphrase);
+
+            // let pass = keys::pwhash(passphrase);
```
cloudhead, line 266:
```suggestion
```

scripts/contributor-workflow-setup.sh
```diff
@@ -66,7 +66,10 @@ mkdir --parents tmp/root
 banner "MAINTAINER"
 ###################
 
-rad auth --init --name cloudhead --passphrase cloudhead
+export RAD_PASSPHRASE=cloudhead
+rad auth --init --name cloudhead
+# Alternatively, the passphrase can be passed via stdin:
+# cloudhead | rad auth --init --name cloudhead --stdin
```
cloudhead, line 72: Perhaps that's cleaner for the script? I think it would be `echo cloudhead | rad auth ...` - not sure if you need the `-n` flag as well..

common/src/test.rs
```diff
@@ -29,6 +29,7 @@ pub mod setup {
 
     pub fn lnk_home() -> Result<(), BoxedError> {
         env::set_var(LNK_HOME, env::current_dir()?.join("lnk_home"));
+        env::set_var("RAD_PASSPHRASE", USER_PASS);
```
cloudhead, line 32: Perhaps introducing a constant for the env var name to avoid typos.

scripts/contributor-workflow-setup.sh
erak: Just checked, and `-n` is not needed.

terminal/src/io.rs
```diff
@@ -310,6 +302,47 @@ pub fn secret_input_with_confirmation() -> SecUtf8 {
     )
 }
 
+pub fn secret_stdin() -> Result<SecUtf8, anyhow::Error> {
+    let mut input: Zeroizing<String> = Zeroizing::new(Default::default());
+    std::io::stdin().read_line(&mut input)?;
```
adaszko, line 307: Can't `stdin().read_line()` read into `SecUtf8` type directly, bypassing `Zeroizing`?  `SecUtf8` does everything `Zeroizing` does and more, so maybe we could ditch the `zeroize` dependency altogether?

erak: Mmh, it seems that it should work with the latest version of `secstr`: https://docs.rs/secstr/latest/secstr/struct.SecUtf8.html#method.unsecure_mut. 

Unfortunately, we're on 0.3.2 though and it's not supported it that version.

erak: Opened a PR in `radicle-keystore` that updates secstr to 0.5.x: https://github.com/radicle-dev/radicle-keystore/pull/35

adaszko: Great, thanks!  I Didn't realise it's going to require a version update 😬

cloudhead: Cool, we can fix this in a later PR

erak: Tracked here: https://github.com/radicle-dev/radicle-cli/issues/240

PR #238, opened 2022-08-30
Store APT GPG key in proper place

salim-b:
`apt-key` [is deprecated](https://manpages.debian.org/testing/apt/apt-key.8.en.html):

> Use of apt-key is deprecated, except for the use of apt-key del in maintainer scripts to remove existing keys from the main keyring. If such usage of apt-key is desired the additional installation of the GNU Privacy Guard suite (packaged in gnupg) is required.
>
> apt-key(8) will last be available in Debian 11 and Ubuntu 22.04.

Following Debian's wiki, the [right place](https://wiki.debian.org/DebianRepository/UseThirdParty) to store third-party repo keys is `/usr/share/keyrings/*`, with the `signed-by` option set in the `sources.list` entry.

This change makes the [`apt-key` deprecation warning](https://askubuntu.com/questions/1398344/apt-key-deprecation-warning-when-updating-system) during `apt update` vanish.

Note that the updated installation instructions should also land on the [Get Started](https://radicle.xyz/get-started.html) page (I don't know where these instructions are sourced from exactly). Currently, instructions there are outdated and don't include https://github.com/radicle-dev/radicle-cli/pull/216

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,11 @@ And then to use it:
 
 First, download the package signing key:
 
-    curl https://europe-west6-apt.pkg.dev/doc/repo-signing-key.gpg | sudo apt-key add -
+    curl -fsSL https://europe-west6-apt.pkg.dev/doc/repo-signing-key.gpg | gpg --dearmor | sudo tee /usr/share/keyrings/radicle-archive-keyring.gpg > /dev/null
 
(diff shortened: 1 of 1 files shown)


adaszko (approved):
Great stuff, thank you for the contribution as well as the explanation!

cloudhead:
Merged, thanks!

Issue #239, opened 2022-08-31
Support issue tracking through versioning


slack-coder:
Help end users communicate the rad version they run by communicating whether they are running release builds.  Many end users are using the client built directly from the project's master branch.

Including the target architecture can help too.

Suggestions given:
  - 0.6.1-ab736f (commit hash appended for dev releases)
  - 0.6.1-(current/post/dev)-ab736f (post or dev describe it is after 0.6.1)
  - 0.6.1-ab736f+armv7 (armv7 is the target of the build, the '+' is a semantic versioning compatible way to include build metadata)

Alternative:
 - use development branches, and use master only for releases

Related links
 - https://www.keiruaprod.fr/blog/2020/05/12/include-version-in-rust-binary.html
 - https://semver.org

gsaslis:
Thanks for thinking about this @slack-coder ! I've also been hit by running a different version than the one I was following instructions for at some point, so this would really help ease that pain in the future. 

Just one clarification, considering `0.6.1` is the latest released version right now, shouldn't the examples in the description have the `0.6.2-` prefix, considering that these are pre-release versions we'd be talking about (as per semver #9 convention - https://semver.org/#spec-item-9 ) ?

slack-coder:
> Just one clarification, considering `0.6.1` is the latest released version right now, shouldn't the examples in the description have the `0.6.2-` prefix, considering that these are pre-release versions we'd be talking about (as per semver #9 convention - https://semver.org/#spec-item-9 ) ?

Maybe we should do this?  The problem is we, and most others, don't track what the next release will be early on, whether 0.6.2 or 0.7.0.  The pre-releases tend to be phased in later, but just before the next release.

erak:
Including the commit hash sounds good, but from my understanding this should be added after the `+`. I'd slightly change the proposal by @slack-coder and throw in the following, considering 

1. release builds: `0.6.1-release+commit.<sha>`
2. development builds: `0.6.2-dev+commit.<sha>`, whereas the above was the previous release. In case we want to have nightly builds a some point, `dev` could be easily replaced by `nightly` and that would perfectly fit the naming scheme.



I'm not sure if we need to include the target for now, since that information could be easily retrieved from the user.

erak:
I'll take a quick look a the implementation, such that we _hopefully_ only need decide on and apply the final scheme.

erak:
[...]

PR #242, opened 2022-09-01
Binary versions with build metadata

erak:
Closes https://github.com/radicle-dev/radicle-cli/issues/239.

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -1,0 +1,21 @@
+use std::process::Command;
+
+fn main() {
+    let hash = Command::new("git")
+        .arg("rev-parse")
+        .arg("--short")
+        .arg("HEAD")
+        .output()
+        .ok()
+        .and_then(|output| {
+            if output.status.success() {
+                String::from_utf8(output.stdout).ok()
+            } else {
+                None
+            }
+        })
+        .unwrap_or_else(|| String::from("unknown"));
+
+    println!("cargo:rustc-env=GIT_HEAD={}", hash);
+    println!("cargo:rustc-rerun-if-changed=.git/HEAD");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4401,7 +4401,7 @@ dependencies = [
 
 [[package]]
 name = "rad-account"
-version = "0.1.0"
+version = "0.7.0-dev"
 dependencies = [
  "anyhow",
  "ethers",
@@ -4412,10 +4412,10 @@ dependencies = [
 
 [[package]]
 name = "rad-anchor"
-version = "0.1.0"
+version = "0.7.0-dev"
 dependencies = [
  "anyhow",
- "coins-bip32 0.6.0",
+ "coins-bip32 0.3.0",
  "colored 1.9.3",
  "ethers",
  "lexopt",
@@ -4431,7 +4431,7 @@ dependencies = [
 
 [[package]]
 name = "rad-auth"
-version = "0.1.0"
+version = "0.7.0-dev"
 dependencies = [
  "anyhow",
  "assay",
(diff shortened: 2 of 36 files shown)


slack-coder:
How about removing the 'vergen' package when things are finalized?  It works, but introducing a new dependency comes at a cost.

erak:
Indeed, `vergen` comes with the cost of additional dependencies. Rebuilding this mechanism from scratch also comes with some cost though. We'd need to call a git subcommand, get the latest revision and, I think that's the tricky part, consistently create the correct Cargo config, such that the environment variable ends up being available to `rad`.

slack-coder:
:+1: I can help out with this once we're settled on the requirements

cloudhead:
See: https://github.com/radicle-dev/radicle-client-services/blob/master/build.rs

slack-coder:
Here's the patch for bumping the version to [0.7.0-dev](https://github.com/slack-coder/radicle-cli/commit/84d8623413109df496c7623a97fa4f5d2e2244eb)

erak:
@slack-coder I picked your commit, thanks! @cloudhead Together with that and the snippet you posted, the version is now being reported as e.g. `rad 0.7.0-dev+cb2c115` for development builds. If it's a release build (version does not contain `-dev`), the sha is not added.

cli/build.rs
```diff
@@ -0,0 +1,21 @@
+use std::process::Command;
+
+fn main() {
+    let hash = Command::new("git")
```
pinkforest, line 4: this is going to be flaky and will break things. it is better to have it in automated builds. build.rs is useful e.g. with protoc for code generation but not for managing build metadata.

pinkforest:
fwiw - 

It's probably worthwhile to start using CHANGELOG.md using "standard" keepchangelog https://keepachangelog.com/en/1.0.0/

e.g. see example use: https://github.com/khonsulabs/bonsaidb/blob/main/CHANGELOG.md

And align on https://semver.org/spec/v2.0.0.html use. 

As to the -dev etc. suffix(es) it's best generated in -nightly, -pr etc. automated builds rather than off manifest which should reflect the "base version"

erak:
> fwiw -
> 
> It's probably worthwhile to start using CHANGELOG.md using "standard" keepchangelog https://keepachangelog.com/en/1.0.0/
> 
> e.g. see example use: https://github.com/khonsulabs/bonsaidb/blob/main/CHANGELOG.md

Oh, yes. We should add a changelog. I've filed an issue tracking this: https://github.com/radicle-dev/radicle-cli/issues/249

> 
> And align on https://semver.org/spec/v2.0.0.html use.
> 
> As to the -dev etc. suffix(es) it's best generated in -nightly, -pr etc. automated "tagged" release-builds rather than off manifest / build.rs which should reflect the "base version" at master / main or the expected future release.

I agree with the base version here, but was under the impression that people were leaning towards this solution.

> 
> e.g. look nightly release builder:
> 
> https://github.com/pinkforest/test-auditable-releases
> 
> It makes nightly releases based on matrix using git tags pushing to timestmaped nightly ver.
> 
> https://github.com/pinkforest/test-auditable-releases/releases/tag/cargo-auditable%40nightly-20220830-0324
> 
> I have a separate repo for radicle-nightly but I'm tuning some of the target triplets that I need for testing:
> 
> https://github.com/pinkforest/rad-release/releases/tag/radicle-cli%40nightly-20220908-0511
> 
> When the CHANGELOG.md is using standardised form typical release tooling has pattern to extract it out automatically.

Cool, that's very valuable! I'm in favor of providing nightly builds. But as mentioned above, it should be possible to distinguish local builds as well. What would be your proposal here?

pinkforest:
>  fear that starting the conversation around nightly builds will block this from being done

It's already built...

> we'll lose CI for now.

They are portable workflows, not really, you can run them off github.

slack-coder:
@pinkforest check the [issue](https://github.com/radicle-dev/radicle-cli/issues/239) for the PR, it may be a good place to follow up so as not to fragment the discussion.

slack-coder:
```
rad --version
# rad 0.7.0-dev+10dfad2
```
I like it :)

PR #256, opened 2022-09-23
rm: Remove profile with peer id

erak:
### Motivation

The intended usage of `rad rm <urn | profile-id>` (in conjunction with `rad auth`) was a bit unclear as the concepts of profiles and identities require deeper technical knowledge.

With this PR `rad rm` needs to be called with a `peer-id` instead of a `profile-id`. This change is also inline with the upcoming protocol update that makes increased usage of Peer IDs.  

Some users got the internal error message `missing field 'subject'` when passing a URN that does not point to a project. This is now also handled more gracefully.

### DIFF
--- a/rm/src/lib.rs
+++ b/rm/src/lib.rs
@@ -6,9 +6,9 @@ use std::str::FromStr;
 use anyhow::anyhow;
 
 use librad::git::Urn;
+use librad::PeerId;
 
 use radicle_common::args::{Args, Error, Help};
-use radicle_common::profile::ProfileId;
 use radicle_common::{keys, profile, project};
 use radicle_terminal as term;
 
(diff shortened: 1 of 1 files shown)


rm/src/lib.rs
```diff
@@ -108,64 +110,70 @@ impl Args for Options {
...
+        }
+        Object::User(peer_id) => {
+            let mut profiles = profile::list()?;
+            if storage.peer_id() != peer_id {
+                profiles.retain(|p| match profile::read_only(p) {
+                    Ok(s) => s.peer_id() == peer_id,
+                    Err(_) => false,
+                });
+
+                if let Some(other) = profiles.last() {
```
FintanH, line 147: I think you could combine `retain` + `last` by doing:
```rust
if let Some(other) = profiles.find(|p| match profile::read_only(p) {
  Ok(s) => s.peer_id() == peer_id,
  Err(_) => false,
}) => ...
```

FintanH (commented):
Just passing by :)

rm/src/lib.rs
cloudhead: Yes, also you could use `map` to not have to re-fetch the `profile::read_only` below

cloudhead (approved):

PR #258, opened 2022-09-25
auth: Use peer id

### LINKED ISSUE #255: Request - Change rad auth command to display name + URN
## Issue

I have multiple identities setup, such that running `rad auth` shows this:
```
(base) bordumb@bordumbs-MacBook-Pro ~ % rad auth
Your active profile is b172f291-f9ab-4990-b94e-738796026f50
  792d1c65-4dcf-4d56-b296-27446491c90d
  364755b3-e777-44c2-a040-06b595a44cbf
* b172f291-f9ab-4990-b94e-738796026f50
```

The problem is that I actually have no clue what any of these URNs are pointing to.
They aren't human readable.

## Proposal

Show both the URN and the name, such that running `rad auth` would display this:
```
(base) bordumb@bordumbs-MacBook-Pro ~ % rad auth
Your active profile is b172f291-f9ab-4990-b94e-738796026f50 (bordumb)
  792d1c65-4dcf-4d56-b296-27446491c90d (testBordumb)
  364755b3-e777-44c2-a040-06b595a44cbf (testBordumb2)
* b172f291-f9ab-4990-b94e-738796026f50 (bordumb)
```

Not too dissimilar from the way it is displayed when changing profiles:

image

erak:
### Motivation

This PR switches from `profile-id` to `peer-id` as the main handle for identities. It makes the `profile-id` less prominent but leaves it in the output for debugging purposes. 

Fixes #255.

It's also supposed to work together with https://github.com/radicle-dev/radicle-cli/pull/256.

### Implementation

```
$ rad auth

Your active identity is hynrwhj53gznypueg4jz7ya9kus81k36zpr8oijdp68bji41zsj85w (user1)
  hynrwhj…1zsj85w (user1)
* hydod3x…7jrhywg (user2)
```
->
```
🌱 Authenticating as hydod3x36munjbjua6ek8kyztr43dotfopo9pwjmnqifcrrp7jrhywg (user2)

ok Profile da7c2aa9-ff79-4e48-bdd8-b0552188025c activated
** Warning: Adding your radicle key to ssh-agent...
ok Passphrase · ********
ok Unlocking...
ok Radicle key added to ssh-agent

```


erak:
Additional note: the user-facing wording should also be more in line with https://docs.radicle.xyz/using-radicle/identity where `identity` the primarily used term.

terminal/src/identity.rs
```diff
@@ -0,0 +1,63 @@
...
+        self.shorten = true;
+        self
+    }
+
+    pub fn style(mut self) -> Self {
+        self.style = true;
+        self
+    }
+
+    pub fn print(self) -> String {
```
cloudhead, line 39: Perhaps this can be moved to be an instance of `fmt::Display`, that way it doesn't have to be called explicitly.

terminal/src/identity.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    pub fn shorten(mut self) -> Self {
+        self.shorten = true;
+        self
+    }
+
+    pub fn style(mut self) -> Self {
+        self.style = true;
+        self
+    }
```
cloudhead, line 37: For methods that move and return `self`, I tend to prefer using adjectives, to communicate the fact that it isn't 'doing' (mutating) anything to `self`. So for example the above would be `short` and `styled`.

terminal/src/identity.rs
```diff
@@ -0,0 +1,63 @@
...
+/// `username` depending on the configuration.
+///
+/// If short display is requested, `peer_id` is printed in its compact format
+/// e.g. `hynddpk…uf4qwge`. If `styled` is requested, `peer_id` and `username`
+/// are printed using the terminal's styled formatters.
+pub struct Formatter<'a> {
+    profile: &'a Profile,
+    shorten: bool,
+    style: bool,
+}
```
cloudhead, line 18: This is a nice idea!

terminal/src/identity.rs
erak: Hehe, had that before and switched to verbs. But yeah, agreed!

terminal/src/identity.rs
```diff
@@ -0,0 +1,63 @@
+use librad::profile::Profile;
+
+use radicle_common::fmt;
+use radicle_common::profile;
+
+use crate as term;
+
+/// Identity formatter that takes a profile and displays it `peer_id` and
```
cloudhead, line 8:
```suggestion
/// Identity formatter that takes a profile and displays it as `peer_id` and
```

terminal/src/identity.rs
cloudhead: Haha cool. Yeah another example is when doing verification: `Foo::verify(&self) -> bool` but `Foo::verified(self) -> VerifiedFoo`

terminal/src/identity.rs
erak: Glad you like it! I introduced this because I think in the future we could potentially share formatting code with the TUI, which works on https://docs.rs/tui/0.19.0/tui/text/struct.Spans.html.

I'd want to see how we could extract shareable components that are independent from how they're written into the buffer.

cloudhead (approved):

PR #261, opened 2022-09-28
contributor-workflow: Use Peer ID for persons

slack-coder:
Fix script to reflect recent changes to the auth command.  It now requires Peer IDs in place of Profile IDs.


erak (dismissed):
Ah, I missed this in my PR. Thanks!

PR #272, opened 2022-11-04
Next Generation CLI tooling

xphoniex:


common/src/profile.rs
```diff
@@ -20,108 +22,22 @@ use crate::keys;
...
         hint: "To setup your radicle profile, run `rad auth --init`.",
     };
 
-    match Profile::active(&home()) {
-        Ok(Some(profile)) => Ok(profile),
-        Ok(None) => Err(not_active_error.into()),
+    match Profile::load() {
+        Ok(profile) => Ok(profile),
+        //Ok(None) => Err(not_active_error.into()),
         Err(_) => Err(error.into()),
```
xphoniex, line 41: Thoughts re this?

terminal/src/display.rs
```diff
@@ -39,23 +37,17 @@ impl<'a> Identity<'a> {
 
 impl<'a> fmt::Display for Identity<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let read_only = profile::read_only(self.profile).map_err(|_| fmt::Error)?;
-        let username = profile::name(Some(self.profile)).map_err(|_| fmt::Error)?;
-        let username = format!("({})", username);
+        //TODO(dave): remove this unwrap
+        let profile = Profile::load().unwrap();
```
xphoniex, line 41: what should the error be here in case profile can't be loaded?

terminal/src/io.rs
```diff
@@ -3,9 +3,7 @@ use std::str::FromStr;
 
 use zeroize::Zeroizing;
 
-use librad::crypto::keystore::pinentry::SecUtf8;
-use librad::crypto::BoxedSigner;
-use librad::profile::Profile;
+use secstr::SecUtf8;
```
xphoniex, line 6: previously, we were using `librad::crypto::keystore::pinentry::SecUtf8;` but using `secstr::SecUtf8` compiles just fine. hope this didn't break anything.

auth/src/lib.rs
```diff
@@ -98,26 +99,23 @@ impl Args for Options {
...
-            anyhow::bail!("you may not specify a peer id when initializing a new identity");
+    if options.init || profile.is_err() {
+        if options.node_id.is_some() {
+            anyhow::bail!("you may not specify a node id when initializing a new identity");
         }
         init(options)
     } else {
-        authenticate(&profiles, options, ctx)
+        authenticate(&profile.unwrap(), options, ctx)
     }
```
cloudhead, line 119: We can get rid of `--init` and simplify this whole function, since you can't init a profile if you already have one!

cloudhead: Also see `radicle-tools/rad-auth` for how to check that the profile doesn't exist.

auth/src/lib.rs
```diff
@@ -43,7 +44,7 @@ pub struct Options {
     pub active: bool,
     pub stdin: bool,
     pub name: Option<String>,
-    pub peer_id: Option<PeerId>,
+    pub node_id: Option<NodeId>,
```
cloudhead, line 47: This option no longer makes sense, since there's only one node id.

common/src/profile.rs
cloudhead: There's a NotFound error that you can match on to return a different message

xphoniex: `args::Error` has these 3 variants: `Help`, `Usage`, and `WithHint` ([source](https://github.com/radicle-dev/radicle-cli/blob/master/common/src/args.rs))

